PR #14, opened 2019-04-03
Salami beatles

magdalenafuentes:
Opening this pull_request to discuss the code, still work in progress


mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+
+def _load_sections(sections_path, annotators):
+    sections_data = []
+    for a in range(len(annotators)):
+        times, secs = [], []
+        for f in ['uppercase.txt', 'lowercase.txt']:
+            if annotators[a]:
+                file_path = os.path.join(sections_path, 'parsed', 'textfile'+str(a+1)+'_'+f)
```
rabitt: `file_path = os.path.join(sections_path, 'parsed', 'textfile{}_{}'.format(str(a+1), f))`

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,185 @@
...
+    global SALAMI_METADATA
+    SALAMI_METADATA = _load_metadata(data_home=data_home)
+
+
+def cite():
+    cite_data = """
+===========  MLA ===========
+Smith, Jordan Bennett Louis, et al. 
+"Design and creation of a large-scale database of structural annotations." 
+ISMIR. Vol. 11. 2011.
```
rabitt: I think it's not the best citation - i think it's treating ISMIR as a journal not a conference.

mir_dataset_loaders/utils.py
```diff
@@ -40,17 +40,19 @@ def validator(dataset_index, data_home):
...
-                if track_id not in missing_files.keys():
-                    missing_files[track_id] = []
-                missing_files[track_id].append(local_path)
-            # validate that the checksum matches
-            elif md5(local_path) != checksum:
-                if track_id not in invalid_checksums.keys():
-                    invalid_checksums[track_id] = []
-                invalid_checksums[track_id].append(local_path)
+            print(filepath)
+            if not filepath == None:  # TODO: needed for empty entries salami, @rabitt ok with this?
```
rabitt: yes! nitpick `if filepath is not None:`

rabitt:
@magdalenafuentes your build is failing because you need to merge your branch with master (the circleci config was updated since your version, so it's failing)

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,194 @@
...
+                    sections_data.append(SectionsData(np.array(times)[:-1],
+                                                      np.array(times)[1:], np.array(secs)))
+                else:
+                    times, secs = None, None
+                    sections_data.append(None)
+            else:
+                times, secs = None, None
+                sections_data.append(None)
+
+    return sections_data
```
rabitt: to make it more clear that `sections_data` is a list of different `SectionData` objects (I was confused at first), maybe instead call it something like `all_annotators_section_data` or something like that.

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,134 @@
...
+     'chords',
+     'key',
+     'sections',
+     'title',
+      'artist']
+)
+
+def download():
+    raise NotImplementedError(
+        "Unfortunately the Beatles dataset is not available for download.")
```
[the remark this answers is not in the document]
magdalenafuentes: Ok. What about annotations? If they're available should download them but not the audio? Or is that the case of `ikala` and I should simply print a message too?

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,134 @@
...
+        track_data['sections'][0].split('/')[-1][:-4],
+        'The Beatles'
+    )
+
+def _load_beats(beats_path):
+    if not os.path.exists(beats_path):
+        return None
+
+    # TODO: fix new points, ignored now - @rabbit, should we ignore, fix or keep those values with no label?
+    with open(beats_path) as fhandle:
```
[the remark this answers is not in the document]
magdalenafuentes: There are different delimiters depending on the file :/

What do you think about this?

```    beat_times, beat_positions = [], []
    with open(beats_path, 'r') as f:
        reader = csv.reader(f)
        for line in reader:
            line = line[0].replace('\t', ' ')
            beat_times.append(float(line.split(' ')[0]))
            beat_positions.append(line.split(' ')[-1]) ```

File: mir_dataset_loaders/beatles.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (rabitt) [reply mir_dataset_loaders/beatles.py]: good question... if the annotations are freely available, I'd say download them (see the way I did it in orchest.download()), and have a message at the end of the download that says that the audio isn't available, and but you have it here's where to put it.

File: mir_dataset_loaders/salami.py [hunk 7]
```diff
@@ -0,0 +1,194 @@
...
+
+SALAMI_INDEX = json.load(open(SALAMI_INDEX_PATH, 'r'))
+SALAMI_METADATA = None
+
+
+SalamiTrack = namedtuple(
+    'SalamiTrack',
+    ['track_id',
+     'sections_annotator_1_lowercase',
+     'sections_annotator_1_uppercase',
```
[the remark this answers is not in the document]
magdalenafuentes: It refers to segmentation hierarchies, they are different annotations. We should keep both!

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,194 @@
...
+                if os.path.exists(file_path):
+
+                    with open(file_path, 'r') as fhandle:
+                            reader = csv.reader(fhandle, delimiter='\t')
+                            for line in reader:
+                                times.append(float(line[0]))
+                                secs.append(line[1])
+
+                    sections_data.append(SectionsData(np.array(times)[:-1],
+                                                      np.array(times)[1:], np.array(secs)))
```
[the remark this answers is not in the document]
magdalenafuentes: The difference was in the `End` label which is an instant, included begin and end time the same for that label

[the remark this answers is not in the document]
rabitt: Should the return then instead be `np.array(secs[:-1])`?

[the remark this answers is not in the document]
magdalenafuentes: I'm keeping the last section label, just assigning it the same start and end (so all arrays are of same length)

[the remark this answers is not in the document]
rabitt: I think I'm still missing something - I don't see from your code how the arrays `np.arrays(times)[:-1]` and `np.array(secs)` can be the same length, when the lists `times` and `secs` are the same lengths.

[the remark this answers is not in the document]
magdalenafuentes: At the end removed the last label (which had no duration and corresponded to the label 'End'). Also remove sections with duration == 0, otherwise it breaks `mir_eval`.

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,134 @@
...
+        sections_data,
+        track_data['sections'][0].split('/')[-1][:-4],
+        'The Beatles'
+    )
+
+def _load_beats(beats_path):
+    if not os.path.exists(beats_path):
+        return None
+
+    # TODO: fix new points, ignored now - @rabbit, should we ignore, fix or keep those values with no label?
```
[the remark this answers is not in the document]
magdalenafuentes: Fixed

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,194 @@
...
+      'annotator_1_time',
+      'annotator_2_time',
+      'broad_genre',
+      'genre']
+)
+
+
+def download():
+    raise NotImplementedError(
+        "Unfortunately the Salami dataset is not available for download.")
```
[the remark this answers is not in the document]
magdalenafuentes: Done.

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,194 @@
...
+
+    if track_id in SALAMI_METADATA.keys():
+        track_metadata = SALAMI_METADATA[track_id]
+    else:
+        # annotations with missing metadata
+        track_metadata = {'source': None, 'annotator_1_id': None, 'annotator_2_id': None, 'duration_sec': None,
+                          'title': None, 'artist': None, 'annotator_1_time': None, 'annotator_2_time': None,
+                          'class': None, 'genre': None}
+
+    annotations_dir = os.path.join(data_home, 'salami-data-public-master','annotations')
```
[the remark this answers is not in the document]
magdalenafuentes: Clarified in the downloader. Let me know if you have doubts.

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+     'chords',
+     'key',
+     'sections',
+     'title']
+)
+
+
+def download(data_home=None, clobber=False):
+    save_path = get_save_path(data_home)  # TODO: @rabitt have to do this so it work with data_home always, better idea?
+    save_path = get_save_path(os.path.join(save_path, BEATLES_DIR, 'annotations'))
```
rabitt: I'd say remove this line, and then when you do the untar, do:
```
beatles_annotations_path = os.path.join(save_path, BEATLES_DIR, 'annotations')
if not os.path.exists(beatles_annotations_path):
    os.makedirs(beatles_annotations_path)
untar(download_path, beatles_annotations_path, cleanup=True)
```

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+import os
+import csv
+
+from . import BEATLES_INDEX_PATH
+from .utils import (get_save_path, get_local_path, validator, BeatData, ChordData, SectionData, KeyData,
+                    download_from_remote, RemoteFileMetadata, untar)
+
+BEATLES_DIR = 'Beatles'
+BEATLES_INDEX = json.load(open(BEATLES_INDEX_PATH, 'r'))
+BEATLES_ANNOT_REMOTE = RemoteFileMetadata(  # TODO: @rabbit this is not metadata but the annotations, do especific data type?
```
rabitt: I think the name `BEATLES_ANNOT_REMOTE` is totally ok, if this is what you're asking :)

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+            If you have the Beatles dataset, place the contents into a folder called
+            Beatles with the following structure:
+                > Beatles/
+                    > annotations/
+                    > audio/
+            and copy the Beatles folder to {}
+        """.format(save_path))
+
+
+def validate(data_home, silence):
```
rabitt: `def validate(data_home=None):` 
for now, ignore the silence parameter

rabitt: remove it below as well

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+    missing_files, invalid_checksums = validator(BEATLES_INDEX, data_home, silence=silence)
+    return missing_files, invalid_checksums
+
+
+def track_ids():
+    return list(BEATLES_INDEX.keys())
+
+
+def load(data_home=None, silence=False):
```
rabitt: same comment about silence - just remove the parameter everywhere. we'll fix all of the silencing issues in #17

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+    beat_times, beat_positions = [], []
+    with open(beats_path, 'r') as f:
+        dialect = csv.Sniffer().sniff(f.read(1024))
+        f.seek(0)
+        reader = csv.reader(f, dialect)
+        for line in reader:
+            beat_times.append(float(line[0]))
+            beat_positions.append(line[-1])
+            if line[-1] == 'New Point':
+                print(beats_path)
```
rabitt: I'm guessing these prints are for debugging? if so, remove :)

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+            start_times.append(float(line[0]))
+            end_times.append(line[1])
+            sections.append(line[3])
+
+    section_data = SectionData(np.array(start_times), np.array(end_times), np.array(sections))
+
+    return section_data
+
+
+def fix_newpoint(beat_positions):
```
rabitt: make this a hidden function - `_fix_newpoint`

mir_dataset_loaders/beatles.py
magdalenafuentes: OK

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+    return beat_data
+
+
+def _load_chords(chords_path):
+    if chords_path is None or not os.path.exists(chords_path):
+        return None
+
+    start_times, end_times, chords = [], [], []
+    with open(chords_path, 'r') as f:
+        dialect = csv.Sniffer().sniff(f.read(1024))
```
rabitt: 🎉

mir_dataset_loaders/beatles.py
magdalenafuentes: Just asking if you want to be extra consistent with data types and want me to create a `RemoteFileAnnotations` (which will be the same as `RemoteFileMetadata` with another name)

mir_dataset_loaders/beatles.py
```diff
@@ -0,0 +1,203 @@
...
+
+def _load_key(key_path):
+    if key_path is None or not os.path.exists(key_path):
+        return None
+
+    start_times, end_times, keys = [], [], []
+    with open(key_path, 'r') as f:
+        reader = csv.reader(f, delimiter='\t')
+        for line in reader:
+            if 'Key' in line[2]:
```
rabitt: `if line[2] == 'Key':`?

mir_dataset_loaders/beatles.py
magdalenafuentes: ok, so it print's invalid/missing files every time you load dataset?

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+from .utils import (get_local_path, validator, SectionData, get_save_path, download_from_remote,
+                    unzip, RemoteFileMetadata)
+
+SALAMI_INDEX = json.load(open(SALAMI_INDEX_PATH, 'r'))
+SALAMI_METADATA = None
+SALAMI_DIR = 'Salami'
+SALAMI_ANNOT_REMOTE = RemoteFileMetadata(
+    filename=os.path.join(SALAMI_DIR, 'salami-data-public-master.zip'),
+    url='https://github.com/DDMAL/salami-data-public/archive/master.zip',
+    checksum='b01d6eb5b71cca1f3163fae4b2cd4c61')  # TODO: @rabbit this are the annotations, do especific data type?
```
rabitt: remove comment

mir_dataset_loaders/beatles.py
magdalenafuentes: saw comment below

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+     'annotator_1_time',
+     'annotator_2_time',
+     'broad_genre',
+     'genre']
+)
+
+
+def download(data_home=None, clobber=False):
+    save_path = get_save_path(data_home)
+    save_path = get_save_path(os.path.join(save_path, SALAMI_DIR))
```
rabitt: same comment as above

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+            If you have the Salami dataset, place the contents into a folder called
+            Salami with the following structure:
+                > Salami/
+                    > salami-data-public-master/
+                    > audio/
+            and copy the Salami folder to {}
+        """.format(save_path))
+
+
+def validate(data_home, silence):
```
rabitt: rm silence parameter here and below

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+    missing_files, invalid_checksums = validator(SALAMI_INDEX, data_home, silence=silence)
+    return missing_files, invalid_checksums
+
+
+def track_ids():
+    return list(SALAMI_INDEX.keys())
+
+
+def load(data_home=None, silence=True):
```
rabitt: rm silence

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+                if os.path.exists(file_path):
+
+                    with open(file_path, 'r') as fhandle:
+                            reader = csv.reader(fhandle, delimiter='\t')
+                            for line in reader:
+                                times.append(float(line[0]))
+                                secs.append(line[1])
+                    times, secs = np.array(times), np.array(secs)
+                    # remove sections with length == 0
+                    times_ = np.delete(times, np.where(np.diff(times) == 0))
```
rabitt: I think naming a variable with a trailing underscore means something specific in python, at least by convention. How about instead:
`times_` --> `times_revised`
and for consistency do the same with `secs`
`secs_revised`

mir_dataset_loaders/salami.py
```diff
@@ -0,0 +1,210 @@
...
+
+    return all_annotators_section_data
+
+
+def _load_metadata(data_home):
+
+    metadata_relative_path = get_local_path(data_home, os.path.join(SALAMI_DIR,
+                                                                    'salami-data-public-master',
+                                                                    'metadata', 'metadata.csv'))
+    metadata_path = get_local_path(
```
rabitt: I think there's a bug here - `metadata_relative_path` will start with data_home already. i'm guessing you don't need `metadata_path` at all, just `metadata_relative_path`

mir_dataset_loaders/utils.py
```diff
@@ -86,8 +87,30 @@ def validator(dataset_index, data_home, silence=False):
     ['start_time', 'end_time', 'lyric', 'pronounciation']
 )
 
+SectionData = namedtuple(
+    'SectionsData',
```
rabitt: fix the names here and for `BeatData`/`ChordData` to match e.g. 
```
SectionData = namedtuple(
    'SectionData',
    ...
)

File: mir_dataset_loaders/utils.py [hunk 27]
```diff
@@ -159,8 +182,7 @@ def download_from_remote(remote, data_home=None, clobber=False):
         os.path.join(MIR_DATASETS_DIR, remote.filename) if data_home is None
         else os.path.join(data_home, remote.filename)
     )
-
-    if not os.path.exists(download_path) or clobber:
+    if not os.path.exists(download_path) or clobber:  # TODO: @rabitt checking for the right file?
```
rabitt: If the file we're about to download already exists, we don't want to overwrite it.

mir_dataset_loaders/salami.py
magdalenafuentes: indeed, they're the same path, removing one

mir_dataset_loaders/utils.py
magdalenafuentes: Ok, but we don't check if there's an unzipped version of the files already, right? so if we run `download` and there's no `.zip` file (because it was cleaned), we'll overwrite the unzipped files. Is that correct?

rabitt:
👍

rabitt:
🎉 🇺🇾

Issue #23, opened 2019-04-03
Add flag to silence validator


rabitt:
`utils.validator()` prints all issues by default. Add an optional flag to skip printing, default should print.

drubinstein:
It may be better to use the `logging` module and use that instead here.

rabitt:
Fixed in #26

PR #26, opened 2019-04-03
Add silence flag to validation

drubinstein:
For #23 

I added a logging function that takes an optional silence flag. It may be better to instead use the python `logging` module and let the user set the level.


rabitt:
Looks great to me. Fine with me to switch to logging if it's possible to have the equivalent of the `slience` flag, if we call the function in different contexts.

rabitt:
Closes #23

PR #33, opened 2019-04-05
python 2.7 and __init__ fixes

andreasjansson:


andreasjansson:
Should fix all the setup issues

PR #35, opened 2019-04-09
Test utils

mmscibor:
Added tests to `utils.py`. I have omitted the `log_message`, which should likely be rethought, as well as `load_json_index` - which I'm not sure if getting `CWD` is the correct approach. 

This should cover most of the tests though.


tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+                   expected_missing,
+                   expected_inv_checksum):
+    index_path = os.path.join("tests/indexes", test_index)
+    with open(index_path) as index_file:
+        test_index = json.load(index_file)
+
+    missing_files, invalid_checksums = utils.validator(test_index, "tests/resources/")
+
+    assert expected_missing == len(missing_files)
+    assert expected_inv_checksum == len(invalid_checksums)
```
drubinstein: Any reason why you can't compare against the expected missing files and invalid checksums and not just the length?

mmscibor: Good point, will make this more thorough

tests/test_utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+from mirdata import utils
+
+import json
+import pytest
+
+if sys.version_info.major == 3:
+    builtin_module_name = 'builtins'
+else:
+    builtin_module_name = '__builtin__'
```
mmscibor: @drubinstein maybe a better way to do this?

drubinstein: `from future import builtins`

drubinstein:
Can you add a test dataset loader. Should that be a separate or combined with three test indexes?

PR #36, opened 2019-04-11
expose beatles and salami

drubinstein:
These weren't exposed in __init__.py causing me troubles. exposing them here.


drubinstein:
@rabitt I'm not an authorized user so I can't merge :(

PR #38, opened 2019-04-12
Fix dataset name in README

rabitt:
remove `mdl` from examples, fix broken links


magdalenafuentes:
what about the folder name [here](https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/utils.py#L23)? Remains the same?

drubinstein:
I think that's fine. Only problem would be on windows, but I doubt you're targeting that?

Issue #17, opened 2019-04-03
Check if dataset is available locally


rabitt:
If a dataset is "available" on disk, we should know it. There are three possible states for a given dataset:
1) exists and valid (the `.validate()` function returns empty dictionaries)
2) exists and not valid (at least one file in the dataset index exists, but some may be missing or have invalid checksums)
3) does not exist (no files in the dataset exist locally)

Let's indicate states (1) and (2) by files called `_VALIDATED` and `_INVALID.json` which live in the dataset folder, e.g. `Orchset/_VALIDATED`.

1) when calling `.download()`, first check if `_VALIDATED` exists. If it does not (or if `clobber=True`) download the dataset / print instructions to download. Otherwise, print something that indicates that the dataset already exists and is valid.
2) In `utils.validator()` if `_VALIDATED` exists, don't run validation. Add a flag to the function to force run even if `_VALIDATED` exists.
3) At the end of `utils.validator()`, create an empty `_VALIDATED` file is run successfully without missing/invalid files. If every file in the index is missing, print something that tells the user the dataset is not available locally. If some files are missing/some are invalid, create `_INVALID.json` which saves the dictionaries of invalid/missing files.

drubinstein:
Why have validated files? Wouldn't it be easier to assume that if it exists, it is valid? If it is ever invalid we delete the directory and tell the user?

magdalenafuentes:
When validating checksums, what happens if the local dataset is in a compressed format, e.g. flac, mp3?

rabitt:
@drubinstein 

> Why have validated files? 

There are a few reasons, the main one being speed. For the upcoming issues #21 and #22 , they'll require having this information. It could be computed when those functions are run, but validating that the checksums are equal is slow to be run each time, especially as the datasets get bigger. 

> Wouldn't it be easier to assume that if it exists, it is valid?

Hard no on this one. One of the driving points for this library is reproducibility/consistency of datasets. It's unfortunately often the case that people will have different versions of the data. We introduced the checksums to make sure, file by file, that a local version of a dataset is consistent with an "official" one.

> If it is ever invalid we delete the directory and tell the user?

If the data were easily available I'd say yes, but very often the data can be precious and hard to get again, so I don't want to be deleting user's files.

rabitt:
@magdalenafuentes 

> When validating checksums, what happens if the local dataset is in a compressed format, e.g. flac, mp3?

[...]

PR #43, opened 2019-04-23
validated/invalid files

drubinstein:
For #17 

I added functions and tests to create the `VALIDATED` and `INVALID.json` files. I've added a pretty good test coverage hopefully via unit tests and tried it out on the beatles datasets with what I think are the correct results (the `.wav`, `.txt` and `.lab` files are missing), which then show up in the json file.

We can merge this in and then adapt the other datasets later or we can do it all in here. Until then. Enjoy.

### DIFF
--- a/mirdata/beatles.py
+++ b/mirdata/beatles.py
@@ -4,14 +4,13 @@ from collections import namedtuple
 import numpy as np
 import os
 import csv
+import json
 
-from .utils import (get_save_path, get_local_path, validator, BeatData,
-                    ChordData, SectionData, KeyData, load_json_index,
-                    download_from_remote, RemoteFileMetadata, untar)
+import mirdata.utils as utils
 
 BEATLES_DIR = 'Beatles'
-BEATLES_INDEX = load_json_index("beatles_index.json")
-BEATLES_ANNOT_REMOTE = RemoteFileMetadata(
+BEATLES_INDEX = utils.load_json_index("beatles_index.json")
+BEATLES_ANNOT_REMOTE = utils.RemoteFileMetadata(
     filename='The Beatles Annotations.tar.gz',
     url='http://isophonics.net/files/annotations/'
         'The%20Beatles%20Annotations.tar.gz',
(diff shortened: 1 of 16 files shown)


mirdata/beatles.py
```diff
@@ -4,14 +4,13 @@
 import numpy as np
 import os
 import csv
+import json
 
-from .utils import (get_save_path, get_local_path, validator, BeatData,
-                    ChordData, SectionData, KeyData, load_json_index,
-                    download_from_remote, RemoteFileMetadata, untar)
+import mirdata.utils as utils
```
mmscibor: Was this the `pytest` issue? This is probably the better way to do these imports anyway

tests/test_beatles.py
```diff
@@ -0,0 +1,86 @@
...
+    assert len(os.listdir(data_home)) == 0
+
+
+def test_download_missing_files(tmpdir, mocker):
+    data_home = str(tmpdir)
+    save_path = utils.get_save_path(data_home)
+    beatles_annotations_path = os.path.join(save_path, beatles.BEATLES_DIR)
+
+    mock_validated = mocker.patch.object(utils, 'check_validated')
+    mock_validated.return_value = False
```
mmscibor: If you are using the same mock across multiple tests, extract them to a 

```
@pytest.fixture
def mock_validated(mocker):
  return mocker.patch.object(utils, 'check_validated')

...

def test_download_missing_files(tmpdir, mocker, mock_validated):
  # etc
```

It's a nice way to, using injection only, guarantee that something doesn't get invoked but rather gets mocked.

mirdata/beatles.py
drubinstein: Yes it was. Also, if we want a better reason why, I talked to a friend at google and he said their guidelines is to always import at the module level.

tests/test_beatles.py
drubinstein: Ooh I was looking for a mechanism to do this. 👍

mirdata/utils.py
```diff
@@ -257,3 +260,28 @@ def load_json_index(filename):
...
+    open(os.path.join(annotations_path, VALIDATED_FILE_NAME), 'a').close()
+
+
+def create_invalid(annotations_path, missing_files, invalid_checksums):
+    with open(os.path.join(annotations_path, INVALID_FILE_NAME), 'w') as f:
+        json.dump({'missing_files': missing_files,
+                   'invalid_checksums': invalid_checksums}, f, indent=2)
+
+
+def clobber_all(remote, annotations_path, data_home=None):
```
rabitt: Nit - `annotations_path` --> `downloaded_content_path` or something similar. This should basically be the unzipped file contents right?

mirdata/beatles.py
```diff
@@ -30,28 +29,45 @@
...
+        print("""
+            Unfortunately the audio files of the Beatles dataset are not available
+            for download. If you have the Beatles dataset, place the contents into
+            a folder called Beatles with the following structure:
+                > Beatles/
+                    > annotations/
+                    > audio/
+            and copy the Beatles folder to {}
+        """.format(save_path))
+        utils.create_invalid(beatles_annotations_path, missing_files, invalid_checksums)
```
rabitt: Can we instead put the `utils.create_invalid` and `utils.create_validated` functions inside `utils.validator`? And `utils.validator` can also first run `utils.check_validated`.

drubinstein: I was thinking about this, but I decided not to because I saw extra logic in the beatles dataset with the print statement and thought that if we shouldnt be writing these files if other datasets do extra with the missing files (what if they have retry logic for example?)

I also felt that it was weird for the validator to take upon itself extra work and require both data_home and the annotations path and make it a little bit trickier to test.

However, doing this would save me a lot of time refactoring all the datasets...

mirdata/utils.py
drubinstein: 👍

mirdata/beatles.py
rabitt: hmm true. Maybe the print statement can be conditioned on the output of `utils.validator`?

drubinstein: Update after in-person conversation. We will move the logic into validator.

drubinstein:
@rabitt Please re-review. I made the changes we discussed.

mirdata/ikala.py
```diff
@@ -41,8 +38,22 @@
...
+                          data_home)
+    if utils.check_validated(dataset_path):
+        print("""
+                The {} dataset has already been validated.
+                If you feel this is a mistake please rerun and set clobber to true.
+                """.format(IKALA_DIR))
+        return
+
+    validate(dataset_path, data_home)
     print("""
```
rabitt: This should only print if validate "fails".

rabitt: Same comment for the other loaders.

drubinstein: updated.

mirdata/utils.py
```diff
@@ -71,6 +74,12 @@ def validator(dataset_index, data_home, silence=False):
                         invalid_checksums[track_id] = []
                     invalid_checksums[track_id].append(local_path)
 
+    return missing_files, invalid_checksums
+
+
+def validator(dataset_index, data_home, dataset_path, silence=False):
+    missing_files, invalid_checksums = check_index(dataset_index, data_home, silence)
```
rabitt: before running `check_index`, first call `check_validated`, and skip running if it's valid.

drubinstein: should be changed now.

Issue #34, opened 2019-04-08
Regularly check if links for downloading data are online


magdalenafuentes:
Just discussed with @rabitt about this. We are using several links to download data, we should check regularly if they are `online`. Is it possible to write a test on that?

If a link goes `offline`, one idea is to write a function that checks links when calling download, and gives a `temporarily offline` message or so. What do you think?

tkell:
Hmm, there are a few ways to do this!

- We could check in CircleCI if the files still exist ... but that is a bit odd because our build would not pass if someone else removes a file.
- We could add a script that makes a HEAD request to each file and confirms that they're there.
- We could modified `download_from_remote` to check for a `URLError` and get out accordingly.  

What do you think?

rabitt:
I actually think a combination of all three could make sense! (3) seems like a good thing to have in general. Could we do both (2) and (3), and have a separate CircleCI process (i.e. (1)) that runs (2), which checks for (3)?

tkell:
kk!  I will start on this stuff!

tkell:
- [ ] Add a `check_remote` function to `utils.py` that makes a HEAD request and returns True or False
- [ ] Write a test that makes this call for all datasets
- [ ] Automate this test with Circle, but _don't fail the build_ if it fails.

drubinstein:
Can we just put on the README.md some embedded javascript or a shield from https://shields.io/ showing whether or not the link is live?

rabitt:
Update - after discussing with @tkell and @drubinstein we decide we won't regularly check links, but instead, report to the user if a link is broken. If it is, we ask the user to report it on our issues page with the flag `broken-link`.

rabitt:
Fixed in #53

PR #47, opened 2019-04-25
[WIP - DO NOT MERGE] Adds check_remote and tests for the same

tkell:
Don't merge, we might just use Requests!

(But this is the first step of #34, and it is bad because making a HEAD request is totally different in py2 and py3)

### DIFF
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -39,7 +39,7 @@ except ImportError:
         request = Request(url)
         request.get_method = lambda : 'HEAD'
         try:
-            response = urllib2.urlopen(request)
+            response = urlopen(request)
             contents = response.read()
             return response.getcode()
         except HTTPError:
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -17,6 +17,15 @@ try:
     from urllib.request import urlretrieve  # py3
 except ImportError:
     from urllib import urlretrieve  # py2
+try:
+    from urllib.request import Request  # py3
+except ImportError:
+    from urllib import Request  # py2
+try:
+    from urllib.error import HTTPError # py3
+except ImportError:
+    from urllib2 import HTTPError # py2
+
 import zipfile
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


mirdata/utils.py
```diff
@@ -167,6 +176,36 @@ def update_to(self, b=1, bsize=1, tsize=None):
...
+        Named tuple containing remote dataset meta information: url, filename
+        and checksum.
+
+    Returns
+    -------
+    file_exists: bool
+        True if file exists.
+        False otherwise.
+    """
+    req = Request(remote.url, method='HEAD')
```
drubinstein: can make this a bit neater I think
```
try:
  res = urlopen(req)
  return res.status >= 200 and res.status < 300
except HTTPError:
  pass
return False

File: tests/test_utils.py [hunk 2]
```diff
@@ -60,6 +60,29 @@ def test_get_save_path(mocker, tmpdir):
...
+    httpserver.serve_content('File not found!', 404)
+
+    TEST_META = utils.RemoteFileMetadata(
+        filename='remote.wav',
+        url=httpserver.url,
+        checksum=('3f77d0d69dc41b3696f074ad6bf2852f')
+    )
+
+    file_exists = utils.check_remote(TEST_META)
+    assert file_exists == False
```
drubinstein: `assert not file_exists`

tests/test_utils.py
```diff
@@ -60,6 +60,29 @@ def test_get_save_path(mocker, tmpdir):
...
+    httpserver.serve_content(open('tests/resources/remote.wav').read())
+
+    TEST_META = utils.RemoteFileMetadata(
+        filename='remote.wav',
+        url=httpserver.url,
+        checksum=('3f77d0d69dc41b3696f074ad6bf2852f')
+    )
+
+    file_exists = utils.check_remote(TEST_META)
+    assert file_exists == True
```
drubinstein: `assert file_exists`

mirdata/utils.py
```diff
@@ -17,6 +17,15 @@
     from urllib.request import urlretrieve  # py3
 except ImportError:
     from urllib import urlretrieve  # py2
+try:
```
drubinstein: You can probably merge these (and all the try..excepts) from the previous blocks. Or maybe we should just do a check on the system major version and do
if sys.version.major >= 3
   do py3 imports
else
 do py2 imports

tkell:
The Py2 / Py3 stuff here is a total disaster, and getting worse, yep.  Does `requests` have a better cross-version API?  (cc @rabitt, are you opposed to `requests`?)

tkell:
Or `from six.moves import urllib`, maybe, again cc @rabitt?

drubinstein:
You can use requests. It's py2/3 compat https://pypi.org/project/requests/ look in the lower left hand corner.

tkell:
Closing to do things with Requests, then do the reduced version of this;  see #34 for discussion!

PR #49, opened 2019-04-25
Add coveralls to tox build

rabitt:


drubinstein:
We're now using codecov instead. Closing.

PR #53, opened 2019-04-26
move to requests, throw friendly exception when downloading a file fails

### LINKED ISSUE #52: Replace urllib with requests
Requests makes Py2 and Py3 compatibility much easier, though it does mean that we have to write our own function for saving large files to disk.

tkell:
Fixes #52 and #34!

- Replaces the try/catch for imports with a single import for `requests`.  (We would have needed another try/catch for the `HTTPError`, which `requests` gives us for free)
- Creates our own `download_large_file` function to replace `urlretrieve`, which is the one thing requests does _not_ do out of the box.
- Updates `download_from_remote` to throw an HTTPError if things fail, with the discussed friendly error message.
- Updates tests accordingly.

I am not sure if `remote.filename` is the correct thing to format the string with.  @rabitt?  Also cc @drubinstein for review.

### DIFF
--- a/mirdata/ikala.py
+++ b/mirdata/ikala.py
@@ -9,13 +9,9 @@ from __future__ import print_function
 from collections import namedtuple
 
 import csv
+import os
 import librosa
 import numpy as np
-import os
-try:
-    from urllib.request import urlretrieve  # py3
-except ImportError:
-    from urllib import urlretrieve  # py2
 
 import mirdata.utils as utils
 
@@ -177,11 +173,10 @@ def _reload_metadata(data_home):
 
 
 def _load_metadata(data_home):
-
     id_map_path = utils.get_local_path(
         data_home, os.path.join(IKALA_DIR, "id_mapping.txt"))
     if not os.path.exists(id_map_path):
-        urlretrieve(ID_MAPPING_URL, filename=id_map_path)
+        utils.downlad_large_file(ID_MAPPING_URL, id_map_path)
 
     with open(id_map_path, 'r') as fhandle:
         reader = csv.reader(fhandle, delimiter='\t')
--- a/mirdata/ikala.py
+++ b/mirdata/ikala.py
@@ -176,7 +176,7 @@ def _load_metadata(data_home):
     id_map_path = utils.get_local_path(
         data_home, os.path.join(IKALA_DIR, "id_mapping.txt"))
     if not os.path.exists(id_map_path):
-        utils.downlad_large_file(ID_MAPPING_URL, id_map_path)
+        utils.download_large_file(ID_MAPPING_URL, id_map_path)
 
     with open(id_map_path, 'r') as fhandle:
         reader = csv.reader(fhandle, delimiter='\t')
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -10,15 +10,8 @@ import os
 import json
 import tarfile
 import shutil
-try:
-    from urllib.request import urlopen  # py3
-except ImportError:
-    from urllib2 import urlopen  # py2
-try:
-    from urllib.request import urlretrieve  # py3
-except ImportError:
-    from urllib import urlretrieve  # py2
 import zipfile
+import requests
 from tqdm import tqdm
 
 MIR_DATASETS_DIR = os.path.join(os.getenv("HOME", "/tmp"), "mir_datasets")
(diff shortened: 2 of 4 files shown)


mirdata/ikala.py
```diff
@@ -177,11 +173,10 @@ def _reload_metadata(data_home):
 
 
 def _load_metadata(data_home):
-
     id_map_path = utils.get_local_path(
         data_home, os.path.join(IKALA_DIR, "id_mapping.txt"))
     if not os.path.exists(id_map_path):
-        urlretrieve(ID_MAPPING_URL, filename=id_map_path)
+        utils.downlad_large_file(ID_MAPPING_URL, id_map_path)
```
drubinstein: Downlad looks like a typo

mirdata/utils.py
```diff
@@ -184,6 +177,17 @@ def update_to(self, b=1, bsize=1, tsize=None):
         self.update(b * bsize - self.n)
 
 
+def download_large_file(url, download_path, callback=None):
+    response = requests.get(url)
+    response.raise_for_status()
+    with open(download_path, 'wb') as handle:
+        for block in response.iter_content(1024):
+            handle.write(block)
+            if callback:
```
drubinstein: Could just default callback to lambda : pass. Is there any reason why we need this here. Is anyone using callbacks?

mirdata/utils.py
```diff
@@ -217,8 +221,16 @@ def download_from_remote(remote, data_home=None, clobber=False):
...
-                remote.url, filename=download_path, reporthook=t.update_to)
+            try:
+                download_large_file(remote.url, download_path, t.update_to)
+            except requests.exceptions.HTTPError:
+                error_msg = 'mirdata failed to download {}! \
+                             Please try again in a few minutes. \
+                             If this error persists, please raise an issue at \
+                             https://github.com/mir-dataset-loaders/mirdata, \
+                             and tag it with "broken-link". \
+                             '.format(remote.filename)
```
drubinstein: Maybe misinterpreting this but if you want a one line print you can also do
'foo'
'bar'
'baz'

And it'll join all the strings so you don't need the backslashes

mirdata/utils.py
```diff
@@ -217,8 +221,16 @@ def download_from_remote(remote, data_home=None, clobber=False):
...
+                raise IOError(error_msg)
```
drubinstein: Maybe create a custom error. Maybe use one of these instead https://docs.python.org/3/library/exceptions.html#ConnectionError

mirdata/utils.py
mmscibor: Generally nice to have callbacks available in anything that is async, although I am not entirely sure how async is handled in python (i.e. is there a way to return a promise?). I think this is fine, in case someone somewhere does want it.

mirdata/utils.py
mmscibor: Why not just re-raise the `HTTPError` with the appropriate message? That's what is happening here

mirdata/utils.py
```diff
@@ -184,6 +177,17 @@ def update_to(self, b=1, bsize=1, tsize=None):
         self.update(b * bsize - self.n)
 
 
+def download_large_file(url, download_path, callback=None):
```
rabitt: is a "large file" treated any differently than a small file? If not, I'd just call this function "download_file"

mirdata/utils.py
rabitt: the best thing would be the dataset name, but that would require an extra function argument just to print... how about make the error message generic and say "... failed to download the dataset!..."

mirdata/ikala.py
tkell: Fixed!

mirdata/utils.py
tkell: @drubinstein, I put it there so we could update `DownloadProgressBar` with each chunk!  Happy to take it out, myself

@mmscibor, this is not async - the callback fires when each block is written.

mirdata/utils.py
tkell: It is!  `download_file_in_chunks_ would be the correct version - can change if you want?

mirdata/utils.py
tkell: Doing `"""` !

mirdata/utils.py
drubinstein: That... Is a good reason! Keep it. I still believe in my comment about the default.

requirements.txt
```diff
@@ -1,6 +1,7 @@
-coveralls>=1.7.0
 six
 testcontainers
+coveralls>=1.7.0
```
drubinstein: Something happened here. Why was coveralls put back in. We don't use it. Also please change requests to >=

tkell: Hmm, I just moved it around: 

```
-coveralls>=1.7.0
 six
 testcontainers
+coveralls>=1.7.0
+requests==2.21.0
```
I feel like @rabitt added it, but am unsure!

drubinstein: Ah. I thought I removed it. Guess I didn't... I'll do that in another pr I suppose.

Issue #54, opened 2019-04-26
Increase test coverage


rabitt:
Increase the test coverage.

For each of the loaders, write simple tests to check the accuracy of the loaders
- [ ] test_beatles.py (move the current tests to the generic tests below)
- [x] test_ikala.py
- [x] test_medleydb_melody.py
- [x] test_medleydb_pitch.py
- [ ] test_salami.py (move the current tests to the generic tests below)
- [x] test_orchset.py

Can we create a "generic" single test module which tests all of the common functions for each loader?
- [ ] test_download.py
- [ ] test_validate.py
- [ ] test_cite.py

rabitt:
Working on this!

PR #55, opened 2019-04-30
Add tests for Salami

tkell:
Working on it! 

Will eventually contribute to #54, pushing for reference.

⛵️ 🚜 🚋

### DIFF
--- a/mirdata/salami.py
+++ b/mirdata/salami.py
@@ -112,7 +112,7 @@ def load_track(track_id, data_home=None):
             'annotator_1_time': None, 'annotator_2_time': None, 'class': None,
             'genre': None
         }
-    salami_path = get_local_path(data_home, SALAMI_DIR)
+    salami_path = utils.get_local_path(data_home, SALAMI_DIR)
     annotations_dir = os.path.join(salami_path,
                                    'salami-data-public-master', 'annotations')
     annotators = [any(SALAMI_INDEX[track_id]['annotator_1_uppercase']),
--- a/tests/test_salami.py
+++ b/tests/test_salami.py
@@ -1,9 +1,11 @@
 from __future__ import absolute_import
 
 import os
-
 import pytest
-
+try:
+    from mock import MagicMock  # py2
+except ImportError:
+    from unittest.mock import MagicMock  # py3
 from mirdata import salami, utils
 from tests.test_utils import (mock_validated, mock_download, mock_unzip,
                               mock_validator, mock_clobber_all)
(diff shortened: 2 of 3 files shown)


tkell:
cc @rabitt, @drubinstein, and @mmscibor for review!

mirdata/salami.py
```diff
@@ -109,7 +109,7 @@ def load_track(track_id, data_home=None):
             'annotator_1_time': None, 'annotator_2_time': None, 'class': None,
             'genre': None
         }
-    salami_path = get_local_path(data_home, SALAMI_DIR)
+    salami_path = utils.get_local_path(data_home, SALAMI_DIR)
```
mmscibor: Is there a `from utils import get_local_path` line that we should remove?

tests/test_salami.py
```diff
@@ -107,3 +104,86 @@ def test_validate_valid(dataset_path, mocker, mock_validator):
...
+def test_load_track_wrong_metadata():
+    with pytest.raises(EnvironmentError):
+        salami.SALAMI_METADATA = {'data_home': 'test'}
+        salami.load_track('10', 'wrong_data_home')
+
+
+def test_load_track_missing_metadata(mocker):
+    data_home = 'fake-data-home'
+    salami.SALAMI_METADATA = {'data_home': data_home}
+    salami.load_sections = mocker.MagicMock(return_value=['a', 'b', 'c', 'd'])
```
mmscibor: This would be the more `mocker` way to do it:

```
mock_load_sections = mocker.patch.object(salami, 'load_sections')
mock_load_sections.return_value = ['a', 'b', 'c', d', 'e']
```

I think with what you have, the state you assign to `salami.load_sections` will hang around for other tests, so you may get some unexpected test failures.

If you do it the way I've described above, the mock will only exist for the scope of this test.

tests/test_salami.py
```diff
@@ -107,3 +104,86 @@ def test_validate_valid(dataset_path, mocker, mock_validator):
...
+def test_load_track_with_metadata(mocker):
+    track_metadata = {
+        'source': 'source', 'annotator_1_id': 'aid-1', 'annotator_2_id': 'aid-2',
+        'duration_sec': 60, 'title': 'title', 'artist': 'artist',
+        'annotator_1_time': None, 'annotator_2_time': None, 'class': 'class',
+        'genre': 'genre'
+    }
+    data_home = 'fake-data-home'
+    salami.SALAMI_METADATA = {'data_home': data_home, '10': track_metadata}
+    salami.load_sections = mocker.MagicMock(return_value=['a', 'b', 'c', 'd'])
```
mmscibor: Same comment as above. Since you are using this mock twice, you can also do something like this:

```
@pytest.fixture
def mock_load_sections():
  return mocker.patch.object(salami, 'load_sections')

...

def test_load_track_with_metadata(mock_load_sections) :
  ...

   mock_load_sections.return_value = ['a' ... 'e']
```

mmscibor: Note that this will also allow you to use a different mock between tests that is correctly scoped and you don't have to worry about things breaking between tests.

mirdata/salami.py
tkell: Nope - it was a mistake, not an import style error!

tests/test_salami.py
tkell: Done!

tests/test_salami.py
tkell: Done!

tests/test_salami.py
```diff
@@ -11,6 +11,10 @@
 def mock_validate(mocker):
     return mocker.patch.object(salami, 'validate')
 
+@pytest.fixture
+def mock_load_sections(mocker):
+  return mocker.patch.object(salami, '_load_sections')
```
mmscibor: I thought it was `load_sections` not `_load_sections`?

mmscibor: Disregard, looking in the code I see now that it is `_load_sections` - that's awesome, looks like we fixed a bad test

tkell:
@mmscibor @rabitt, could you either make me an Authorized User :tm:, or merge this?  Merci!

drubinstein:
Can you rebase off master first? I'll squash and merge afterwards. I'm not a fan of the "update branch" button.

PR #64, opened 2019-06-21
Suggestion: rename download argument "clobber" to "force_overwrite"

andreasjansson:
Just a suggestion.

Method signatures should be self-explaining if possible. "clobber" is not immediately obvious.


lostanlen:
Agreed. I'm not a native English speaker, and i didn't know what `clobber` meant until today. `force_overwrite` is more self-explanatory.

PR #65, opened 2019-06-21
format everything with black; add black check to ci

andreasjansson:


drubinstein:
Can you please add versioning to the requirements you moved to setup.py?

PR #95, opened 2019-07-20
Investigating `black` mismatches

tkell:
My local run of `black -S` changes things that appear to be OK with the CI version of `black -S`, so I am trying things out with this PR to see what it does.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist = py27,py36,py37,black
 
-[testenv]
+[testenv:py27]
 passenv = CIRCLECI CIRCLE_* CI_PULL_REQUEST
 extras =
   tests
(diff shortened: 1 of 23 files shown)


tkell:
Hello @andreasjansson and @rabitt and @drubinstein, it looks like `black` in CI is not working?  The latest commit in this PR is wrong, and the CI passes it.

Will look into this and update here!

tkell:
Update:  for some reason, Tox was not inheriting commands / running commands properly.  I've made a more-verbose-but-working `tox.ini`, and runs `black -S` so that the new section passes.

Let me know what you think!

tox.ini
```diff
@@ -1,16 +1,33 @@
...
+passenv = CIRCLECI CIRCLE_* CI_PULL_REQUEST
+extras =
+  tests
+  docs
+commands =
+  pytest --cov-report=xml --cov=mirdata {toxinidir}/tests/
+
+[testenv:py37]
+passenv = CIRCLECI CIRCLE_* CI_PULL_REQUEST
+deps = black
```
drubinstein: we probably dont need this if [testenv:black] is doing it instead.

drubinstein:
My guess is what you're seeing is the files being formatted for a specific version of python. My guess is that we should choose a specific version of python and use the `-t` flag of black to fix that. By default it does per file auto detection and who knows what different versions of black will do. 

Otherwise, we should just fix it to the most recent version of python. I'm a bigger fan of that honestly. We just need to do a black-bump PR every new release and that's easy enough.

tox.ini
tkell: 💯

tkell:
My theory is that the default testenv pytest command was overwriting the `black` command, but I agree with you about supplying a version - update coming!

tkell:
Closing this PR as it is out of date and it is easier to re-run `black` that do this rebase.

PR #97, opened 2019-07-22
Fix when annotations are missing and test

magdalenafuentes:


tests/test_salami.py
```diff
@@ -85,6 +85,46 @@ def test_track():
     assert track.__repr__() == repr_string
 
 
+    # Test file with missing annotations
+    track = salami.Track('192', data_home=data_home)
+
+    # test attributes 
+    assert track.source == 'Codaich'
```
rabitt: no issue to have all these (lines 92-103), but if you want to skip these and just test for the section annotations you can!

PR #99, opened 2019-07-22
Common downloader

rabitt:
Just a proposal! If we go with this, I need some help with updating the tests @drubinstein @tkell 😬


rabitt:
@drubinstein before I go any further - is this more or less what you had in mind?

mirdata/web_downloader.py
```diff
@@ -0,0 +1,201 @@
...
+            A list of  RemoteFileMetadata tuples of uncompressed data.
+            If None, there is no uncompressed data to download.
+        info_message (str or None):
+            A string of info to print when this function is called.
+            If None, no string is printed.
+        force_overwrite (bool):
+            If True, existing files are overwritten by the downloaded files.
+
+    """
+    Path(save_dir).mkdir(exist_ok=True)
```
drubinstein: Is there a chance that we will ever have zip, tar and file downloads all in the same dataset? If not, it'd probably be easier to split this up into separate functions depending on the download type and call those functions separately.

mirdata/web_downloader.py
```diff
@@ -0,0 +1,201 @@
...
+        tfile = tarfile.open(tar_path, 'r:gz')
+    else:
+        tfile = tarfile.TarFile(tar_path, 'r')
+    tfile.extractall(save_dir)
+    tfile.close()
+    if cleanup:
+        os.remove(tar_path)
+
+
+def force_delete_all(remote, data_home):
```
drubinstein: I think this could stay in utils since it's not web specific.

mirdata/web_downloader.py
```diff
@@ -0,0 +1,201 @@
...
+        os.remove(zip_path)
+
+
+def download_tar_file(tar_remote, save_dir, force_overwrite, cleanup=False):
+    tar_download_path = download_from_remote(
+        tar_remote, save_dir, force_overwrite)
+    untar(tar_download_path, save_dir, cleanup=cleanup)
+
+
+def untar(tar_path, save_dir, cleanup=False):
```
drubinstein: This could also stay in utils since it's not web download specific.

mirdata/web_downloader.py
```diff
@@ -0,0 +1,201 @@
...
+    return download_path
+
+
+def download_zip_file(zip_remote, save_dir, force_overwrite, cleanup=False):
+    zip_download_path = download_from_remote(
+        zip_remote, save_dir, force_overwrite)
+    unzip(zip_download_path, save_dir, cleanup=cleanup)
+
+
+def unzip(zip_path, save_dir, cleanup=False):
```
drubinstein: same comment about leaving this in utils

drubinstein:
@rabitt I left some comments with my review. This is more or less what I envisioned, just some nitpicking. Having a kitchen sink function is probably not going to scale well if we need to add other web downloaders. Grouping all the web download utilities in the same file is 💯 . I'm wondering if we should just name it downloader. Are you ever downloading from not the web nowadays?

mirdata/web_downloader.py
rabitt: seldom both zip and tar, but often zip + file and tar + file. 

For one simple reason, I'd like to keep one mega function - we're going to test everything in `web_downloader`, we're not going to test each `loader.download()`. If we can have each `loader.download()` be a very thin wrapper on top of `web_downloader.downloader()`, the lack of test coverage isn't as bad as if each `loader.download()` has it's own logic.

mirdata/web_downloader.py
rabitt: true. So far it's only used internally by other functions in web_download - do you still think it should move?

rabitt:
@tkell @drubinstein - would either of you have time to help write tests for the new functions in `download_utils.py`? Not immediately urgent :)

tkell:
@rabitt, yep, but probably not for a week or three!  Or David can do them, I am easy.

PR #115, opened 2019-08-23
fix for guitarset.download()

### LINKED ISSUE #113: guitarset.download doesn't create the right folder structure
everything is unzipped directly to `~/mir_datasets/GuitarSet` instead of into the specific subfolders. This is partially because `download_utils.downloader` by default downloads directly into `data_home`. We should add an option to `RemoteFileMetadata` which optionally specifies a subfolder to create and unzip data into.

tomxi:
This should close #113 
Sorry for the mess the first time around.


tomxi:
OK... help needed to hit these two lines with tests.
Don't know how to do that right now with the mocks.

rabitt:
This looks great, and thanks for making the fix! I had a slightly more general idea for how to handle subfolders - going to push a commit to this branch. Let me know what you think!

rabitt:
@tomxi If this looks good to you i'll merge it! I'm ignoring the patch coverage change - I had to add some lines of untested code in `orchset.download()` which is causing the problem.

PR #119, opened 2019-08-30
Checksum and minor fixes

### LINKED ISSUE #116: beatles.download checksum error
When I run `mirdata.beatles.download()`, everything appears to download as expected, but the checksums don't match. Can someone else run the same and see if you can reproduce it? The checksum of my downloaded file is `c3b7d505e033ea9ff0d7a1d57871f2ee`, vs the expected `62425c552d37c6bb655a78e4603828cc`

magdalenafuentes:
- Fixes  #116 
- Simplifies tar.gz download
- RWC beat positions to int

### DIFF
--- a/mirdata/download_utils.py
+++ b/mirdata/download_utils.py
@@ -220,9 +220,6 @@ def untar(tar_path, cleanup=False):
     cleanup: bool, default=False
         If True, remove tarfile after untarring.
     """
-    # if tar_path.endswith('tar.gz'):
-    #    tfile = tarfile.open(tar_path, 'r:gz')
-    # else:
     tfile = tarfile.open(tar_path, 'r')
     tfile.extractall(os.path.dirname(tar_path))
     tfile.close()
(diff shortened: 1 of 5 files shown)


mirdata/download_utils.py
```diff
@@ -220,10 +220,7 @@ def untar(tar_path, cleanup=False):
     cleanup: bool, default=False
         If True, remove tarfile after untarring.
     """
-    if tar_path.endswith('tar.gz'):
-        tfile = tarfile.open(tar_path, 'r:gz')
-    else:
-        tfile = tarfile.TarFile(tar_path, 'r')
+    tfile = tarfile.open(tar_path, 'r')
```
rabitt: Do you know why the original code was necessary/why it's not necessary now?

magdalenafuentes: Not completely, I was using that code to open `tar.gz` files but it breaks with `Beatles` so I changed back to the simple code.

rabitt:
@magdalenafuentes this is ready to merge, just update the version before you do!

PR #124, opened 2019-09-06
make all subdirectories

### LINKED ISSUE #121: `dataset.download` fails if `data_home` doesn't exist
We need to make the top level directory if it doesn't exist in `downloader.download()`

rabitt:
Fixes #121

### DIFF
--- a/mirdata/download_utils.py
+++ b/mirdata/download_utils.py
@@ -13,11 +13,6 @@ import requests
 from requests.exceptions import HTTPError
 from tqdm import tqdm
 
-try:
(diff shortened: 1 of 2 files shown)


drubinstein:
Why do you need this? The `mkdir(exist_ok=True)` should make the directory whether it exists or not.

rabitt:
The existing code makes a directory if it doesn’t exist but fails it it has to make multiple directories (e.g if the “mir_datasets” directory doesn’t exist and t tries to make “mir_datasets/Orchset”.

Issue #111, opened 2019-08-23
Support multichannel wav files


rabitt:
Guitarset `mirdata/guitarset.py` has 6-channel wav files. Librosa supports this, but I think we need to get librosa 0.7.0 working in the build environment. Right now we're raising `NotImplementedError` if someone tries to access the 6-channel audio files - once this is ready we can update the guitarset module.

rabitt:
@tomxi We just got librosa 0.7.0 working and I think it should support multichannel wav files. feel free to give it a shot if you want!

PR #125, opened 2019-09-07
multi channel support w/ GuitarSet

tomxi:
implements #111

Should be a pretty straight forward CR.


tomxi:
addressed #126 as well in this PR.

PR #130, opened 2019-09-27
to JAMS functionality [WIP]

### LINKED ISSUE #85: Jams and annotations consistency
We should write a ‘to_jams’ function to convert data to the jams format, and make clear we don’t test annotations consistency but only that the loaders are doing their job. The ‘to_jams’ function would allow people to check themselves.

### LINKED ISSUE #129: SectionData format a bit different from mir_eval
It has `start_times`, `end_times`, and `sections` attributes. mir_eval expects an array of intervals (which means concatenate start_times and end_times) and an array with sections.

Should we modify the `SectionData` object accordingly?

magdalenafuentes:
function to convert mirdata annotations into jams

When this is done will
* Close #85 
* Close #129


magdalenafuentes:
Now we have:
- a big `jams_converter` function that handles data types and particularities of the datasets
- a obligatory `to_jams()` function to implement in each track, converting whatever makes sense for that dataset

then to use is simply `track.to_jams()`. An example is in `rwc_jazz` for now. Thoughts? @rabitt @tomxi

mirdata/jams_utils.py
```diff
@@ -0,0 +1,65 @@
...
+        beat_data=None,
+        chord_data=None,
+        note_data=None,
+        f0_data=None,
+        section_data=None,
+        key_data=None,
+        artist=None,
+        duration=None,
+        title=None,
+        annotator=None):
```
rabitt: I guess there's going to be more than this as things grow. for now, maybe we should have a `sandbox` variable where we throw any extra information.

mirdata/jams_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    if artist is not None:
+        jam.file_metadata.artist = artist
+    if title is not None:
+        jam.file_metadata.title = title
+    if duration is not None:
+        jam.file_metadata.duration = duration
+
+    # annotations
+
+    # beats
```
rabitt: let's break each of these into separate functinos - input is a list, output is a jams annotation object

mirdata/jams_utils.py
magdalenafuentes: I'm not sure what to do with this, in particular:
1. should we include in the `jams` object **all** the metadata we have per track? 
2. should we have some "mandatory" fields such as `artist` and `duration` and leave the rest to a sandbox, or put everything in the sandbox directly?

magdalenafuentes:
In the `rwc_pop` there's voice-instrument activity annotations, which are `EventData` type. What's the best way to translate that to `jams`? @rabitt @tomxi 

Also @bmcfee might have suggestions to this PR.

mirdata/jams_utils.py
```diff
@@ -0,0 +1,147 @@
...
+        annotator (dict or None):
+            A dict describing the annotator (e.g. {name: annotator_1}
+        sandbox (dict or None):
+            A dict with extra information related to the track.
+
+    """
+
+    jam = jams.JAMS()
+
+    # metadata
```
rabitt: maybe we can have metadata be a dictionary input, e.g.
{ 'artist': 'asdf', 'duration': 270, 'title': 'asdfasdf'}

Then when we create the jams object, we have logic that checks if `jam.file_metadata` has the dictionary key as an attribute (as you did before), and if not adds it to the sandbox.

mirdata/jams_utils.py
```diff
@@ -0,0 +1,147 @@
...
+    return jam
+
+
+def beats_to_jams(beats):
+    jannot_beat = jams.Annotation(namespace='beat')
+    jannot_beat.annotation_metadata = jams.AnnotationMetadata(data_source='mirdata')
+    for t, p in zip(beats[0].beat_times, beats[0].beat_positions):
+        jannot_beat.append(time=t, duration=0.0, value=p)
+    if beats[1] is not None:
+        jannot_beat.sandbox = jams.Sandbox(name=beats[1])
```
rabitt: maybe this fits better in annotation metadata?

mirdata/salami.py
```diff
@@ -27,14 +27,15 @@
...
+import mirdata.jams_utils as jams_utils
 
 INDEX = utils.load_json_index('salami_index.json')
 METADATA = None
 DATASET_DIR = 'Salami'
 ANNOTATIONS_REMOTE = download_utils.RemoteFileMetadata(
-    filename='salami-data-public-master.zip',
-    url='https://github.com/DDMAL/salami-data-public/archive/master.zip',
-    checksum='f88b3455f1f2443458d094f603c0c1c7',
+    filename='salami-data-public-hierarchy-corrections.zip',
```
rabitt: Since we're updating this, we should update the info at the top of the module, to be clear which annotation version we're using!

mirdata/jams_utils.py
```diff
@@ -0,0 +1,147 @@
...
+            A float describing the track duration in seconds.
+        title (str or None):
+            A string describing the track title.
+        annotator (dict or None):
+            A dict describing the annotator (e.g. {name: annotator_1}
+        sandbox (dict or None):
+            A dict with extra information related to the track.
+
+    """
+
```
rabitt: It's annoying to write, but we should do some type checking so that if people input wrong data, it doesn't silently produce garbage results, or error much futher downstream where it's hard to debug. E.g. checking if `beat_data` is either `None` or a list of tuples with the correct types.

tests/test_jams_utils.py
```diff
@@ -0,0 +1,20 @@
...
+
+    beat_data_1 = [(utils.BeatData(np.array([0.2, 0.3]), np.array([1,2])), None)]
+    beat_data_2 = [(utils.BeatData(np.array([0.5, 0.7]), np.array([2,3])), 'beats_2')]
+    beat_data_3 = [(utils.BeatData(np.array([0.0, 0.3]), np.array([1,2])), 'beats_1'),
+                   (utils.BeatData(np.array([0.5, 0.13]), np.array([3, 4])), 'beats_2')]
+    beat_data_4 = (utils.BeatData(np.array([0.0, 0.3]), np.array([1, 2])), 'beats_1')
+
+    assert type(jams_utils.jams_converter(beat_data=beat_data_1)) == jams.JAMS
+
+    # with pytest.raises(TypeError):
```
rabitt: if you add the type checking in `jams_utils` these tests will be super clean to write :)

tests/test_jams_utils.py
```diff
@@ -0,0 +1,20 @@
...
+def test_beats():
+
+    beat_data_1 = [(utils.BeatData(np.array([0.2, 0.3]), np.array([1,2])), None)]
+    beat_data_2 = [(utils.BeatData(np.array([0.5, 0.7]), np.array([2,3])), 'beats_2')]
+    beat_data_3 = [(utils.BeatData(np.array([0.0, 0.3]), np.array([1,2])), 'beats_1'),
+                   (utils.BeatData(np.array([0.5, 0.13]), np.array([3, 4])), 'beats_2')]
+    beat_data_4 = (utils.BeatData(np.array([0.0, 0.3]), np.array([1, 2])), 'beats_1')
+
+    assert type(jams_utils.jams_converter(beat_data=beat_data_1)) == jams.JAMS
+
```
rabitt: check also that the internally stored jams beat data is what we expect it to be

mirdata/jams_utils.py
magdalenafuentes: you mean as an `AnnotationMetadata` object? I think it doesn't have a `name` field. Otherwise I don't know if `jannot_beat.annotation_metadata = jams.Sandbox(name=beats[1])` makes sense?

rabitt: hmm ok, what if we make it a big more flexible. The annotation metadata object has these fields:
`jams.AnnotationMetadata(curator=None, version='', corpus='', annotator=None, annotation_tools='', annotation_rules='', validation='', data_source='')`

what if, like we're doing for track metadata, the beats tuple is now `(BeatData, dict)` where we treat the keys of the dictionary like we do for the track metadata - assigning them to the fields of the annotation_metadata when they exist, and when they dont assigning them to the sandbox?

mirdata/jams_utils.py
```diff
@@ -0,0 +1,257 @@
...
+    jam = jams.JAMS()
+
+    # metadata
+    if metadata is not None:
+        if 'duration_sec' in metadata.keys():  # match name to jams attribute
+            metadata['duration'] = metadata.pop('duration_sec')
+        if 'Title' in metadata.keys():  # match name to jams attribute
+            metadata['title'] = metadata.pop('Title')
+        if 'Artist' in metadata.keys():  # match name to jams attribute
+            metadata['artist'] = metadata.pop('Artist')
```
rabitt: I think this bit doesn't belong here - any logic converting a dataset's specific metadata should probably happen in the dataset's `to_jams` function, and this function can blindly add keys when they match. But, we should probably document which keys are supported.

mirdata/jams_utils.py
```diff
@@ -0,0 +1,257 @@
...
+            if type(lyrics) != tuple:
+                raise TypeError(
+                    'lyrics_data should be a list of tuples, '
+                    + 'but is a list of something else')
+            jam.annotations.append(lyrics_to_jams(lyrics))
+
+    return jam
+
+
+def beats_to_jams(beats):
```
rabitt: for readability, i think the function should take two arguments - beat_data and beat_metadata as input.

rabitt: same for the others below

mirdata/jams_utils.py
```diff
@@ -0,0 +1,257 @@
...
+    jannot_beat = jams.Annotation(namespace='beat')
+    jannot_beat.annotation_metadata = jams.AnnotationMetadata(data_source='mirdata')
+    if beats[0] is not None:
```
rabitt: why do you need to check for `beats[0]` being None? I guess if the beats are none, they shouldn't get passed into the function in the first place right?

mirdata/jams_utils.py
```diff
@@ -0,0 +1,257 @@
...
+        if type(beats[0]) != utils.BeatData:
```
rabitt: the above case will actually be handled here!

mirdata/beatles.py
```diff
@@ -248,7 +254,7 @@ def _load_chords(chords_path):
             chords.append(line[2])
 
     chord_data = utils.ChordData(
-        np.array(start_times), np.array(end_times), np.array(chords)
+        np.array(start_times), np.array(end_times), chords
```
rabitt: while you're at it, want to fix #135 too?

mirdata/medleydb_pitch.py
```diff
@@ -113,6 +114,10 @@ def pitch(self):
     def audio(self):
         return librosa.load(self.audio_path, sr=None, mono=True)
 
+    def to_jams(self):
+        return jams_utils.jams_converter(f0_data=[(self.pitch, None)],
+                                         metadata=self._track_metadata)
```
rabitt: maybe worth making a remapped track_metadata dictionary with some adjusted names, e.g. 'Artist' --> 'artist'

mirdata/salami.py
```diff
@@ -175,6 +179,17 @@ def sections_annotator_2_lowercase(self):
...
 
+    def to_jams(self):
+        return jams_utils.jams_converter(multi_section_data=[(
+                                         [(self.sections_annotator_1_uppercase, 0),
+                                          (self.sections_annotator_1_lowercase, 1)], 'annotator_1'),
+                                         (
+                                         [(self.sections_annotator_2_uppercase, 0),
+                                         (self.sections_annotator_2_lowercase, 1)], 'annotator_2'),
+                                         ],
+                                         metadata={key: self._track_metadata[key]
```
rabitt: overkill no? why not just `metadata=self._track_metadata`?

mirdata/jams_utils.py
magdalenafuentes: Yeah, makes sense. It was a sanity check I added. I'll remove it and stand in the docs that `to_jams()` supports `jams` metadata keys

mirdata/jams_utils.py
magdalenafuentes: mmmm... not sure how to incorporate annotation metadata given the structure it has now, let's chat about it

mirdata/jams_utils.py
magdalenafuentes: could be the case you're missing beats only in a few tracks in a `dataset` and when trying `to_jams()` then it will break

mirdata/jams_utils.py
magdalenafuentes: True!

mirdata/beatles.py
magdalenafuentes: yep, monster PR anyway

mirdata/salami.py
magdalenafuentes: yep

mirdata/jams_utils.py
magdalenafuentes: Actually, if I remove the check, what happens is that if one file has the beats missing it will break. Instead, If I leave it, it will work and return a `jams` object without beats (for that specific file). I think I prefer this latter behavior, what do you think?

magdalenafuentes:
Also fixes #135

mirdata/jams_utils.py
rabitt: Let's leave it for now and open an issue for after this PR is merged!

mirdata/jams_utils.py
rabitt: Same as above - let's punt this for later

CONTRIBUTING.md
```diff
@@ -347,4 +359,4 @@ Bibtex format citations/s here
   a. Include all audio and annotation files for one track of the dataset
   b. For each audio/annotation file, reduce the audio length to a few seconds and remove all but a few of the annotations.
   c. If the dataset has a metadata file, reduce the length to a few lines to make it trival to test.
-2. Test all of the dataset specific code, e.g. the Track object, any of the load functions, and so forth – see the ikala dataset tests (`tests/test_ikala.py`) for reference.
+2. Test all of the dataset specific code, e.g. the Track object, any of the load functions, and so forth ??? see the ikala dataset tests (`tests/test_ikala.py`) for reference.
```
rabitt: Remove the ??? and open an issue :)

tests/test_rwc_jazz.py
```diff
@@ -96,10 +121,20 @@ def test_load_metadata():
...
+        'duration': 167,
         'variation': 'Instrumentation 1',
         'instruments': 'Pf',
     }
+    assert metadata['RM-J044'] == {
+        'piece_number': 'No. 44',
+        'suffix': 'M04',
+        'track_number': 'Tr. 09',
+        'title': 'Joyful, Joyful, We Adore Thee',
+        'artist': 'K’s Band',
```
rabitt: I think this is what's breakign the Circle CI tests - change the apostrophe in K's.

tests/test_rwc_jazz.py
```diff
@@ -96,10 +121,20 @@ def test_load_metadata():
         'track_number': 'Tr. 04',
         'title': 'Crescent Serenade (Piano Solo)',
         'artist': 'Makoto Nakamura',
-        'duration_sec': '02:47',
+        'duration': 167,
         'variation': 'Instrumentation 1',
         'instruments': 'Pf',
     }
+    assert metadata['RM-J044'] == {
```
rabitt: Another one of the tests is saying `KeyError: 'RM-J044'` - maybe this key needs to be added to the partial metadata file in the tests?

PR #134, opened 2019-10-02
Move back from requests to urllib

### LINKED ISSUE #133: Download progress bar doesn't work
We changed something, and the progress bar when running `dataset.download` doesn't work :(

drubinstein:
Fixes #133

### DIFF
--- a/mirdata/download_utils.py
+++ b/mirdata/download_utils.py
@@ -9,9 +9,10 @@ from collections import namedtuple
 import os
 import tarfile
 import zipfile
-import requests
-from requests.exceptions import HTTPError
 from tqdm import tqdm
+from six.moves import urllib
+import six
+import sys
 
 from mirdata.utils import md5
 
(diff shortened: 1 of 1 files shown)


mirdata/download_utils.py
```diff
@@ -129,19 +117,11 @@ def download_from_remote(remote, save_dir, force_overwrite=False):
     if not os.path.exists(download_path) or force_overwrite:
         # If file doesn't exist or we want to overwrite, download it
         with DownloadProgressBar(
-            unit='B', unit_scale=True, miniters=1, desc=remote.url.split('/')[-1]
+            unit='B', unit_scale=True, unit_divisor=1024, miniters=1
         ) as t:
-            try:
-                _download_large_file(remote.url, download_path, t.update_to)
-            except HTTPError:
-                error_msg = """
```
rabitt: What will happen now if there's a dead link? Is there a way to catch that specific error explicitly and keep the error message below?

drubinstein: Updated so we will raise the error with the same message as before.

PR #137, opened 2019-10-22
Speed up import mirdata

### LINKED ISSUE #100: import mirdata is slow
When we import mirdata, each of the loaders are imported, which loads each of the json indexes into memory. No bueno.

We should load the index on the fly or when it's first accessed to speed things up.

rabitt:
Initial idea for how to fix #100 .
WIP, will extend to all loaders if this is the way we decide to go.


drubinstein:
Yea this works. I'd see if there's a way to abstract this out into utils so you dont have to rewrite the init once logic each time or see if you could switch it over to use a singleton instead.

PR #149, opened 2019-11-04
MSD data loader (partial index)

andreasjansson:
A little different from the other loaders since it's such a huge dataset. The `validate` function only validates a 1% sample of the md5 hashes, and there's no `load` function.


mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+class Track(object):
+    """MSD Track class
+
+    Args:
+        track_id (str): Track id of the track
+        data_home (str): Local path where the dataset is stored.
+            If `None`, looks for the data in the default directory, `~/mir_datasets`
+
+    Attributes:
+        Please refer to http://millionsongdataset.com/
```
rabitt: Can you document the top level attributes? `analysis`, `metadata`, etc?

mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+            "msd.Track("
+            + "analysis={analysis}, "
+            + "metadata={metadata}, "
+            + "musicbrainz={musicbrainz})"
+        ).format(
+            analysis=self.analysis, metadata=self.metadata, musicbrainz=self.musicbrainz
+        )
+
+
+class MSDAnalysis(object):
```
rabitt: I know it's a pain but can you add documentation for each of these classes?

mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+
+    # format: 'data/J/D/V/TRJDVIB12903CF9F35.h5'
+    return os.path.join(
+        data_home, "data", track_id[2], track_id[3], track_id[4], track_id + ".h5"
+    )
+
+
+def track_ids(data_home=None):
+    """Generator that yields all MSD track IDs
+
```
rabitt: Add "Args:" to docstring

mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+    Returns:
+        (generator) Track IDs
+    """
+
+    if data_home is None:
+        data_home = utils.get_default_dataset_path(DATASET_DIR)
+
+    for _, _, files in os.walk(data_home):
```
rabitt: I don't like that the output of this function changes depending on what data you have locally.
In the other loaders we're getting the track_ids from the index file. Did you end up including only a subset of the data in the index file? If so, could we store a list of the trackids somewhere, so the user can know what they're missing?

mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+    """Print the reference"""
+
+    cite_data = """
+=========== MLA ===========
+Thierry Bertin-Mahieux, Daniel P.W. Ellis, Brian Whitman, and Paul Lamere.
+The Million Song Dataset. In Proceedings of the 12th International Society
+for Music Information Retrieval Conference (ISMIR 2011), 2011.
+========== Bibtex ==========
+@inproceedings{Bertin-Mahieux2011,
+  author = {Thierry Bertin-Mahieux and Daniel P.W. Ellis and Brian Whitman and Paul Lamere},
```
rabitt: I think the names have to be inverted - like
`author = {Bertin-Mahieux, Thierry and ... `

mirdata/msd.py
```diff
@@ -0,0 +1,373 @@
...
+  title = {The Million Song Dataset},
+  booktitle = {{Proceedings of the 12th International Conference on Music Information
+        Retrieval ({ISMIR} 2011)}},
+  year = {2011},
+  owner = {thierry},
```
rabitt: remove `owner` and `timestamp`

setup.py
```diff
@@ -20,6 +20,8 @@
             'six',
             'jams',
             'requests',
+            'h5py',
```
rabitt: does this get used anywhere?

scripts/make_msd_index.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -eux
+
+MSD_FOLDER=/msd
+
+find $MSD_FOLDER -name *.h5 > msd-index.txt 2>/dev/null
+
+shuf msd-index.txt | head -n10000 > msd-sample.txt
```
rabitt: Can you set a seed/make this deterministic?

mirdata/msd.py
andreasjansson: Yep, will do!

mirdata/msd.py
andreasjansson: Yep, will do!

mirdata/msd.py
andreasjansson: Will fix.

mirdata/msd.py
andreasjansson: I know what you mean, but I think there needs to be a distinction between "small" datasets where the IDs reasonably fit in the repository, and "large" ones like the MSD. The MSD index file, even if we only included filenames, would be ~35MB. If we included hashes as well it'd be well over 100MB, forcing users of mirdata to install git-lfs. I opted for a filesystem-based `track_ids()` function, validated by a probabilistic `validate()` function. It's just a suggestion though, so if you have other ideas I'd happily rewrite this!

scripts/make_msd_index.sh
andreasjansson: Yep!

setup.py
andreasjansson: Actually no, will remove.

andreasjansson:
On your comment about h5 files, MSD doesn't have audio, the dataset consists of a million h5 files.

mirdata/msd.py
rabitt: I agree on the validation side! I just think intuitively a function that gives the dataset's track ids should be deterministic, if possible. Some ideas:
* What about a file that just stores the track ids (no filenames or checksums)?
* I opened an idea for how we deal with the growing size of the `indexes/` folder in #153 - we could do something similar here (download the trackids file from zenodo as part of the download).

andreasjansson: I like that idea! In that case we could even store the full index on Zenodo. If you have bandwidth to download ~500GB of MSD, you can probably download another ~100MB of index. The `validate` method could get an optional `subsample_percent=1.0` (or something) argument, in case you want faster validation.

andreasjansson:
@rabitt I fixed all the issues, except the indexing one. That one seems like a bigger thing to think about, so perhaps we can merge this PR and come back to it once we have consensus on how to deal with large indices?

scripts/make_msd_index.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -eux
```
rabitt: Can you list the dependencies and make the folder name a variable so that it would be easy for someone to rerun this?

setup.py
rabitt: still not used right?

andreasjansson:
Sorry missed those, fixed now @rabitt

lostanlen:
@andreasjansson thank you for your PR. Following your suggestion in #153, we are undertaking the process of migrating mirdata to LFS (Large File Storage). In light of this, we will be able to offer a full index of the MSD, with 100% of the checksums, not 1%. This will also allow to implement a `load` function. However, our migration to LFS will need a force push, and thus invalidate this PR. Therefore, it won't be possible to rebase; instead, you will have to make a new PR to the LFS-enabled version of mirdata.

Are you on board with this?

andreasjansson:
@lostanlen I haven't been following the discussion, but that sounds great! Happy to close this and open a new PR off the new master once that's completed. Unless you prefer this PR to land before the force push?

lostanlen:
no, i think it's best not to merge this PR. In this way, mirdata sticks to the quality standard of "100%-verifiable datasets or bust". ok?

andreasjansson:
Yeah I like that a lot! Ping me once you've migrated to LFS and I'll generate a full index and re-open this PR.

PR #155, opened 2019-11-20
Add GTZAN-Genre dataset

andreasjansson:


mirdata/gtzan_genre.py
```diff
@@ -0,0 +1,154 @@
...
+        self.track_id = track_id
+
+        if data_home is None:
+            data_home = utils.get_default_dataset_path(DATASET_DIR)
+
+        self._data_home = data_home
+        self._track_paths = DATA.index[track_id]
+
+        self.genre = track_id.split(".")[0]
+        self.audio_path = self._track_paths["audio"][0]
```
rabitt: I don't think it's useful to have the relative path exposed explicitly in an attribute - can you merge it into one line:
`self.audio_path = os.path.join(self._data_home, self._track_paths["audio"][0]`

mirdata/gtzan_genre.py
```diff
@@ -0,0 +1,154 @@
...
+        self.abs_audio_path = os.path.join(self._data_home, self.audio_path)
+
+    @property
+    def audio(self):
+        audio, _ = librosa.load(self.abs_audio_path, sr=22050, mono=True)
+        return audio
```
rabitt: Return the sample rate too. Why load it at 22050 and not the audio's native samplerate `sr=None`?

andreasjansson: GTZAN is mono/22.5k. I just made that the default and made `.audio()` a function rather than a property.

PR #170, opened 2020-02-11
document validator + print message in case of success

### LINKED ISSUE #167: utils.validator is undocumented
See: https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/utils.py#L106

This is already a to-do in the code, but i figured it deserved its own issue number.

lostanlen:
Fixes #167

### DIFF
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -62,13 +62,12 @@ def log_message(message, silence=False):
         print(message)
 
 
-def check_index(dataset_index, data_home, silence=False):
+def check_index(dataset_index, data_home):
     """check index to find out missing files and files with invalid checksum
 
     Args:
         dataset_index (list): dataset indices
         data_home (str): Local home path that the dataset is being stored
-        silence (bool)
 
     Returns:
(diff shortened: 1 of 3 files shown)


lostanlen:
Done. This passes tests and has 100% diff hit.
I think this is ready for CR.

mirdata/utils.py
```diff
@@ -103,24 +102,46 @@ def check_index(dataset_index, data_home, silence=False):
 
 
 def validator(dataset_index, data_home, silence=False):
-    """validate.. (todo: what does it do?) """
-    missing_files, invalid_checksums = check_index(dataset_index, data_home, silence)
+    """logs missing files and files with invalid checksums in a dataset index.
```
magdalenafuentes: Maybe we should elaborate this idea a bit more? I'm not sure if a new user will understand it otherwise. Something like "checks the existence and/or validity of files stored locally with respect to the paths and file checksums stored in the reference index, and logs missing or invalid files." Just an idea

lostanlen: thank you. fixed in 5df7c5f

Issue #151, opened 2019-11-05
Project description on pypi is lacking


alastair:
If I view this library on pypi I get a very basic description that doesn't tell me much about the project

> Common loaders for MIR datasets

It'd be nice to extend this with either the readme from the repository, or a different detailed description

lostanlen:
Relatedly, there is no definition of the MIR acronym anywhere in the mirdata package. It might be good to state somewhere (README + PyPI description) that MIR means music information retrieval.

lostanlen:
closed by #171

PR #171, opened 2020-02-12
add project description on PyPI

lostanlen:
fixes ~#36~ #151 (sorry, i initially wrote #36 because i was working on birdvoxdetect at the same time!)

update description, long_description and classifiers


setup.py
```diff
@@ -1,16 +1,30 @@
...
 
+with open('README.md', 'r') as fdesc:
+    long_description = fdesc.read()
+
 if __name__ == '__main__':
     setup(
         name='mirdata',
         version='0.1.1',
-        description='Common loaders for MIR datasets.',
+        description='Data management utilities for reproducible research in music information retrieval (MIR).',
```
magdalenafuentes: I got used to the "common loaders for reproducible...", but this works for me too

lostanlen: It's not clear what "common loaders" mean. "Common"= "widespread" or "common"="compatible"?

lostanlen: @magdalenafuentes note that @rabitt approved this change, so if you're OK with this `description` string, we can resolve the conversation and merge this PR.
This is not an irreversible design decision, so it's fine to GTD and fine-tune later.

lostanlen:
> If I understand well this reads from the README right? What happens if we add the .gif in the README? will this one render it too?

Yes @magdalenafuentes, this hooks into the README and uses it as a README on PyPI.
I think it's OK. PyPI can render GIFs.

One example that comes to mind is tqdm, which has a GIF logo:
https://github.com/tqdm/tqdm

https://pypi.org/project/tqdm/

Does this answer your Q? shall we merge this?

magdalenafuentes:
yes let's merge!

lostanlen:
I set version to `0.2.dev0` in accordance with PEP 440
https://www.python.org/dev/peps/pep-0440/

scikit-learn does the same

lostanlen:
@rabitt this PR is next in line. it explicitly removes py27 support

PR #181, opened 2020-03-05
define version module

### LINKED ISSUE #173: missing __version__
IIUC there is no way to inspect the version of mirdata from inside the module ATM
This is because there is no `__version__` global variable, and `setup.py` relies on a hard-coded constant.
It would be good to have a `version.py` module so containing this constant, and have `setup.py` read from that constant. This is what sklearn and librosa do, among others.

lostanlen:
fixes #173


lostanlen:
triaged for v0.1.2 with approval by rabitt

lostanlen:
imp is deprecated in Python 3
will replace by importlib, which will break py27 support
therefore, triaged to v0.2.0-alpha

v0.1.2 will be the last version of mirdata supporting py27

PR #182, opened 2020-03-05
[MRG] Add MultipitchData

### LINKED ISSUE #101: medleydb_melody melody3 type not in mir_eval format
`mir_eval` expects a ragged array for multipitch outputs, but `melody3` data in `mirdata` is a numpy array.

rabitt:
* Fixes #101 by adding a new data type called `utils.MultipitchData` which is compatible with mir_eval's multipitch.evaluate function.
* Updates `medleydb_melody` to load melody3 data as MultipitchData

### DIFF
--- a/mirdata/medleydb_melody.py
+++ b/mirdata/medleydb_melody.py
@@ -120,7 +120,7 @@ class Track(object):
             + "is_instrumental={}, n_sources={}, "
             + "melody1=F0Data('times', 'frequencies', confidence'), "
             + "melody2=F0Data('times', 'frequencies', confidence'), "
-            + "melody3=F0Data('times', 'frequencies', confidence'))"
+            + "melody3=MultipitchData('times', 'frequencies', confidence'))"
         )
         return repr_string.format(
             self.track_id,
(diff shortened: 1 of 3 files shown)


mirdata/medleydb_melody.py
```diff
@@ -156,12 +156,9 @@ def audio(self):
         return librosa.load(self.audio_path, sr=None, mono=True)
 
     def to_jams(self):
+        # jams does not support multipitch, so we skip melody3
         return jams_utils.jams_converter(
-            f0_data=[
-                (self.melody1, 'melody1'),
-                (self.melody2, 'melody2'),
-                (self.melody3, 'melody3'),
```
lostanlen: i'm not sure how `self.melody3` is stored, but couldn't we write a `jams_converter` for it? it seems to me that JAMS does support multipitch.

Screen Shot 2020-03-05 at 6 51 11 PM

rabitt: yep you're right. Writing the change now

lostanlen: update: actually, it seems impossible, because JAMS is row-major (sources first, time frames later) whereas mir_eval is column-majoe (time frames first, sources later). This causes an inherent ambiguity in the conversion of the ragged array, because there is no deterministic one-to-one mapping between sources.

It would be good to document this limitation of `MultipitchData`

rabitt:
To summarize - we decided to, for now, not write a multipitch_to_jams converter since jams does not have a multipitch namespace, and the multiple pitch_contour annotations would live at the same level of abstraction as the melody1 and melody2 annotations.

Should be ready to merge.

PR #185, opened 2020-03-06
Updating NoteData type to be compatible with mir_eval trancription

### LINKED ISSUE #161: jams_converter ignores note_data
The `jams_utils` module has no `notes_to_jams`, which causes the `jams_converter` kwarg `note_data` to be ignored.

https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/jams_utils.py#L33

magdalenafuentes:
Addresses #161 

[EDIT by @lostanlen: closes #161]

### DIFF
--- a/mirdata/jams_utils.py
+++ b/mirdata/jams_utils.py
@@ -265,8 +265,8 @@ def notes_to_jams(notes):
     -------
     jannot_notes: JAM note_midi annotation object.
     '''
-    jannot_chord = jams.Annotation(namespace='note_midi')
-    jannot_chord.annotation_metadata = jams.AnnotationMetadata(data_source='mirdata')
+    jannot_note = jams.Annotation(namespace='note_hz')
+    jannot_note.annotation_metadata = jams.AnnotationMetadata(data_source='mirdata')
     if notes[0] is not None:
         if type(notes[0]) != utils.NoteData:
             raise TypeError('Type should be NoteData.')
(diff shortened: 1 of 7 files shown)


mirdata/jams_utils.py
```diff
@@ -117,6 +117,20 @@ def jams_converter(
...
+
+    # notes
+    if note_data is not None:
+        if type(note_data) != list:
+            raise TypeError('note_data should be a list of tuples')
+        for notes in note_data:
+            if type(notes) != tuple:
+                raise TypeError(
+                    'note_data should be a list of tuples, '
+                    + 'but is a list of something else'
```
lostanlen: would be nice to say : "but contains a `type(notes)` element`

magdalenafuentes: Sorry I forgot about this one! I don't understand the comment, I guess it could be a list of anything, including  `type(note)`?

lostanlen: @magdalenafuentes lists in Python may contain elements of mixed types.
`[1, 2.0, 'a', (3, 4)]` is perfectly valid as a list.

This is why the phrasing of "X is a list of something else" is a bit misleading. For all we know, some elements in `note_data` may be tuples while others may be non-tuples. From the user's perspective, it's informative to know what sort of non-tuple is in there. The error will be triggered at the first non-tuple encountered.

Does that make sense?

lostanlen:
thank you for this, @magdalenafuentes. I must say, i still don't really like the phrasing of `but it a list of type(notes)`. In Python, lists can contain heterogeneous types. For all we know, some elements in the list may not be of type `type(notes)`. This is why i advocated for the following error message: `"but contains a type(notes) element"` instead. I'm happy to provide a fix in a separate PR.

PR #188, opened 2020-03-06
better support for remote data

### LINKED ISSUE #128: mirdata doesn't work cleanly for datasets not on disk
For the use case where the data files live on a remote machine and are accessed when needed for e.g. for training a model, many of mirdata's assumptions fail:
* validation expects files to be local
* annotation loaders look for files locally
* annotation loading functions are hidden from the user
* `Track` annotation attributes are loaded in the background from disk, expecting files to be present

How can we support this (relatively common) use case cleanly? This will be increasingly important for larger datasets. Some initial ideas:
* expose annotation loading functions to the user
* create functions for validating a single file within a dataset
* better handle errors when accessing `Track` attributes that load files from disk

cc @faroit

rabitt:
Will fix #128 

Done:
- [x] added a `load_audio` function for every dataset
- [x] explicitly defined annotation paths in `Track` objects for every dataset
- [x] made load_annotation type functions public instead of private

Still To do:
- [x] Update/write documentation for new attributes and public functions
- [x] Update tests for all the loaders
- [x] Add an example in the documentation of loading local vs. remote data


mirdata/beatles.py
```diff
@@ -123,6 +127,20 @@ def to_jams(self):
         )
 
 
+def load_audio(audio_path):
```
lostanlen: couldn't this function be shared across datasets, in most cases? like, `from utils import load_audio`; and if not, implement it inside the dataset-specific module

rabitt: Yeah I was thinking the same. I realize it's redundant, but it also feels sort of silly to have a common function that just wraps librosa, which would need to have the same parameters as the librosa function to be generically useful. Right now I'm writing very specific wrappers per dataset on the librosa function that hard codes certain parameters. What do you think?

lostanlen: i'm not sure. i would say: feel free to carry on with your WIP PR and we'll advise later. i can live with it.

mirdata/salami.py
```diff
@@ -89,30 +89,42 @@ def _load_metadata(data_home):
...
+        audio_path (str): TODO
+        broad_genre (str): TODO
+        duration (float): TODO
+        genre (str): TODO
+        sections_annotator1_lowercase_path (str): TODO
+        sections_annotator1_uppercase_path (str): TODO
+        sections_annotator2_lowercase_path (str): TODO
+        sections_annotator2_uppercase_path (str): TODO
+        source (str): TODO
+        title (str): TODO
```
magdalenafuentes: annotator_1_id (str): number that identifies annotator 1
        annotator_1_time (str): time that the annotator 1 took to complete the annotation
        annotator_2_id (str): number that identifies annotator 1 
        annotator_2_time (str): time that the annotator 1 took to complete the annotation
        artist (str): song artist
        audio_path (str): path to the audio file
        broad_genre (str): broad genre of the song
        duration (float): duration of song in seconds
        genre (str): genre of the song
        sections_annotator1_lowercase_path (str): path to annotations in hierarchy level 1 from annotator 1
        sections_annotator1_uppercase_path (str): path to annotations in hierarchy level 0 from annotator 1
        sections_annotator2_lowercase_path (str): path to annotations in hierarchy level 1 from annotator 2
        sections_annotator2_uppercase_path (str): path to annotations in hierarchy level 0 from annotator 2
        source (str): dataset or source of song
        title (str): title of the song

mirdata/salami.py
```diff
@@ -89,30 +89,42 @@ def _load_metadata(data_home):
...
-        sections_annotator_2_lowercase: annotation
+
+    Cached Properties:
+        sections_annotator_1_lowercase (SectionData): TODO
+        sections_annotator_1_uppercase (SectionData): TODO
+        sections_annotator_2_lowercase (SectionData): TODO
+        sections_annotator_2_uppercase (SectionData): TODO
+
+    Properties:
+        audio: TODO
```
magdalenafuentes: Cached Properties:
        sections_annotator_1_lowercase (SectionData): annotations in hierarchy level 1 from annotator 1
        sections_annotator_1_uppercase (SectionData): annotations in hierarchy level 0 from annotator 1
        sections_annotator_2_lowercase (SectionData): annotations in hierarchy level 1 from annotator 2
        sections_annotator_2_uppercase (SectionData): annotations in hierarchy level 0 from annotator 2
    Properties:
        audio: audio file, sample rate

rabitt:
@magdalenafuentes made your changes!

mirdata/__init__.py
```diff
@@ -18,7 +18,7 @@
     'orchset',
     'rwc_classical',
     'rwc_jazz',
-    'rwc_classical',
+    'rwc_popular',
```
lostanlen: oof, good catch.

rabitt:
Merging!

PR #205, opened 2020-04-02
add a base Track class, remove all reprs

### LINKED ISSUE #178: Track base class
Create a base class which Track objects inheret from which implements:
- [ ] `__repr__` in a standard way
- [ ] `.to_jams` as a `NotImplementedError` to force Track objects to have this method

(follow up from #160 )

rabitt:
Implements #178 [EDIT by @lostanlen: closes #178]

All dataset `Track` classes now inherit from `utils.Track`, which has an automated implementation of `__repr__`, removing the need for new loaders to write this themselves 🎉 

It has one peculiarity, which in my opinion ends up being kind of nice -
`__repr__` will break if implemented `Track` properties do not have documentation, and our tests catch that.


lostanlen:
`__repr__ will break if implemented Track properties do not have documentation, and our tests catch that.` 😍

mirdata/utils.py
```diff
@@ -29,11 +29,45 @@
...
+        properties = [v for v in dir(self.__class__) if not v.startswith('_')]
+        attributes = [
+            v for v in dir(self) if not v.startswith('_') and v not in properties
+        ]
+
+        repr_str = "Track(\n"
+
+        for attr in attributes:
+            val = getattr(self, attr)
+            if isinstance(val, str):
```
andreasjansson, line 49: should this be `if not isinstance`?

andreasjansson: if you support py2 still (do you?) you need to check baseinstance for `basestring`

mirdata/utils.py
```diff
@@ -29,11 +29,45 @@
...
 import os
 import json
+import types
 
 
 MIR_DATASETS_DIR = os.path.join(os.getenv('HOME', '/tmp'), 'mir_datasets')
 
 
+class Track(object):
+    def __repr__(self):
```
andreasjansson, line 39: right now you're not truncating strings or anything right? if an object has a really long value it might hang ipython.

mirdata/utils.py
```diff
@@ -29,11 +29,45 @@
 import hashlib
 import os
 import json
+import types
 
 
 MIR_DATASETS_DIR = os.path.join(os.getenv('HOME', '/tmp'), 'mir_datasets')
 
 
+class Track(object):
```
andreasjansson, line 38: i might have put this in `track.py`. `utils.py` is a bit of an antipattern, or at least good to avoid.

mirdata/utils.py
```diff
@@ -29,11 +29,45 @@
...
+
+        for prop in properties:
+            val = getattr(self.__class__, prop)
+            if isinstance(val, types.FunctionType):
+                continue
+
+            if val.__doc__ is None:
+                raise ValueError("{} has no documentation".format(prop))
+
+            val_type_str = val.__doc__.split(':')[0]
```
andreasjansson, line 61: do you validate that docstrings are correctly formatted?

tests/test_beatles.py
```diff
@@ -40,18 +40,6 @@ def test_track():
     assert sr == 44100
     assert audio.shape == (44100 * 2,)
 
-    repr_string = (
```
andreasjansson, line 43: why remove these tests instead of keeping them as regression tests of the superclass repr? wouldn't it be nice if the superclass repr could be backwards-compatible? or is that a bunch of extra work?

mirdata/utils.py
rabitt: I think it's right - I'm formatting strings with visible `"` around the values when printing

mirdata/utils.py
rabitt: nope - should I? if it doesn't have a `:` character somewhere it'll just print the whole docstring, which is ugly but not broken

tests/test_beatles.py
rabitt: They're definitely not backwards compatible - namely getting the property data types printing exactly the same is basically not possible without huge hacks. The previous repr functions also didn't have consistent implementations. 

The main goal of this PR was to make it ligher weight to contribute a new loader. One of the "annoying" things to add was the repr function + tests, since it's a bit tedious to write.

mirdata/utils.py
rabitt: good idea - moving!

mirdata/utils.py
rabitt: good point - will add this

Issue #197, opened 2020-03-12
DALI and GTZAN_genre lack a `to_jams` method


magdalenafuentes:

rabitt:
Also gtazn_genre.to_jams is not implemented

lostanlen:
I will address urgently.
Note that JAMS has a built-in namespace for GTZAN tags
https://github.com/marl/jams/blob/master/jams/schemata/namespaces/tag/gtzan.json

rabitt:
The Gtzan half of this issue is fixed in #212

PR #212, opened 2020-04-06
gtzan_genre.Track.to_jams

### LINKED ISSUE #217: f0s_to_jams is not compliant with pitch_contour namespace
The function `jams_utils.f0s_to_jams` produces a `jam` object whose `Annotation` is marked as `pitch_contour`. However, this `Annotation` is not compliant with the official `pitch_contour` namespace of JAMS. The consequence is that orchset Tracks cannot be exported to JAMS. 

Minimal non-working example:

```python
from mirdata import orchset

X = orchset.load()
keys = list(X.keys())
key = keys[0]
jam = X[key].to_jams()
jam.validate()
```

Result:
```
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
~/miniconda3/lib/python3.7/site-packages/jams/core.py in validate(self, strict)
   1807         try:
-> 1808             schema.VALIDATOR.validate(self.__json_light__, schema.JAMS_SCHEMA)
   1809 

~/miniconda3/lib/python3.7/site-packages/jsonschema/validators.py in validate(self, *args, **kwargs)
    352             for error in self.iter_errors(*args, **kwargs):
--> 353                 raise error
    354 

ValidationError: None is not of type 'number'

Failed validating 'type' in schema['properties']['file_metadata']['properties']['duration']:
    {'minimum': 0.0, 'type': 'number'}

On instance['file_metadata']['duration']:
    None

During handling of the above exception, another exception occurred:

SchemaError                               Traceback (most recent call last)
<ipython-input-2-dbd09b7b0fbe> in <module>
----> 1 jam.validate()

~/miniconda3/lib/python3.7/site-packages/jams/core.py in validate(self, strict)
   1821         except jsonschema.ValidationError as invalid:
   1822             if strict:
-> 1823                 raise SchemaError(str(invalid))
   1824             else:
   1825                 warnings.warn(str(invalid))

SchemaError: None is not of type 'number'

Failed validating 'type' in schema['properties']['file_metadata']['properties']['duration']:
    {'minimum': 0.0, 'type': 'number'}

On instance['file_metadata']['duration']:
    None
```

It appears that the values in the Annotation of `f0s_to_jams` are of type `float`, as they describe the frequency in Hertz.
However, `pitch_contour` values are supposed to be triplets: `(index, frequency, voicing)`.
Note that voicing is boolean ... this is pre-Bittner and Bosch ISMIR 2019 ;)

lostanlen:
This PR implements a JAMS parser for `gtzan_genre`, in the prospect of closing #197 ("DALI and GTZAN Genre lack a to_jams method")

This was quite easy. I am going to write unit tests now.

PS: nevermind that this branch is called `dali-jams`. This is a `gtzan_genre` PR.

EDIT: [rabitt]
This PR now does the following:
* fixes #217 
* adds a test to `test_loaders` which ensures that the output of every `to_jams()` method successfully validates
* the above test broke for nearly every loader, because `duration` was not part of the file_metadata. I've added duration to every `to_jams` method.
* added a new argument to `jams_utils.jams_converter` for gtzan tags.
* changes the convention in `jams_utils` to always us `if not isinstance(...)` over `if type(...) != ...`

### DIFF
--- a/mirdata/gtzan_genre.py
+++ b/mirdata/gtzan_genre.py
@@ -36,6 +36,17 @@ DATASET_REMOTE = download_utils.RemoteFileMetadata(
 
 DATA = utils.LargeData("gtzan_genre_index.json")
 
+VERSION = "1.0"
+
+ANNOTATION_RULES = """
+Unfortunately the database was collected gradually and very early on in my
+research so I have no titles.
+The files were collected in 2000-2001 from a variety of sources including
+personal CDs, radio, microphone recordings, in order to represent a variety of
+recording conditions. Nevetheless I have been providing it to researchers upon
+request mainly for comparison purposes etc. -- George Tzanetakis, 2005
+"""
+
 
 class Track(track.Track):
     """gtzan_genre Track class
--- a/mirdata/gtzan_genre.py
+++ b/mirdata/gtzan_genre.py
@@ -78,7 +78,7 @@ class Track(track.Track):
 
 
 def load_audio(audio_path, sample_rate=22050):
-    """Load a GTzan audio file.
+    """Load a GTZAN audio file.
 
     Args:
         audio_path (str): path to audio file
--- a/mirdata/gtzan_genre.py
+++ b/mirdata/gtzan_genre.py
@@ -14,12 +14,11 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import jams
 import librosa
 import os
 
-import mirdata
 from mirdata import download_utils
+from mirdata import jams_utils
 from mirdata import track
 from mirdata import utils
 
@@ -33,19 +32,7 @@ DATASET_REMOTE = download_utils.RemoteFileMetadata(
     destination_dir="gtzan_genre",
 )
 
-
 DATA = utils.LargeData("gtzan_genre_index.json")
-
-VERSION = "1.0"
-
-ANNOTATION_RULES = """
-Unfortunately the database was collected gradually and very early on in my
-research so I have no titles.
-The files were collected in 2000-2001 from a variety of sources including
-personal CDs, radio, microphone recordings, in order to represent a variety of
-recording conditions. Nevetheless I have been providing it to researchers upon
-request mainly for comparison purposes etc. -- George Tzanetakis, 2005
-"""
 
 
 class Track(track.Track):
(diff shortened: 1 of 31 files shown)


mirdata/gtzan_genre.py
```diff
@@ -73,12 +86,48 @@ def audio(self):
         return load_audio(self.audio_path, sample_rate=22050)
 
     def to_jams(self):
-        """(Not Implemented) Jams: the track's data in jams format"""
-        raise NotImplementedError
+        """Jams: the track's data in jams format"""
+        # Initialize top-level JAMS container
+        jam = jams.JAMS()
```
rabitt, line 91: This should use the [`mirdata.jams_utils.jams_converter`](https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/jams_utils.py#L12) function. If it's missing fields, we should add them to the converter.

lostanlen: @rabitt so i should make a `gtzan_to_jams` function in `jams_utils`? and pass a `gtzan_data` (default `None`) to the `jams_converter`? that sounds overkill to me, but i can live with that.

rabitt:
A caveat - the big change I pushed results in a less complete jams object in the `to_jams` function. Unfortunately, without substantially changing the converter, this is not trivial and I propose we leave this issue for the future.

mirdata/jams_utils.py
```diff
@@ -155,16 +160,28 @@ def jams_converter(
...
+    if tags_data is not None:
+        if not isinstance(tags_data, list):
+            raise TypeError('tags_data should be a list of tuples')
+        for tag in tags_data:
+            if not isinstance(tag, tuple):
+                raise TypeError(
+                    'tags_data should be a list of tuples, '
+                    + 'but contains a {} element'.format(type(tag))
+                )
+            jam.annotations.append(tag_gtzan_to_jams(tag))
```
lostanlen, line 185: `tag_gtzan_to_jams` returns an `Annotation` in the `gtzan` namespace. This `Annotation` will be JAMS-compliant if and only if the `tag` is a GTZAN genre: country, jazz, rock, etc.

This tag converter will not work for other datasets than GTZAN (e.g. MSD, which requires `tag_open`)

rabitt: Yes - this is why I called the method "tag_gtzan_to_jams"

rabitt: I see though - the value is called "tags_data" so it's confusing. I'll update.

PR #231, opened 2020-04-09
[WIP] naming consistency

### LINKED ISSUE #210: Dataset hyphenation
Spotted during CR of #207 

Datasets such as `RWC-Popular`, `RWC-Classical` have a hyphen in their `DATASET_DIR` string, but have a space in the dataset table (#203)

What should be do about this inconsistency?

lostanlen:
closes #210 

This PR adds a new global variable `.name` which is the name of the dataset. In every module, I set the name according to how it was defined at the top of the docstring. I shortened the docstring accordingly to avoid duplicating information.

This PR is backwards compatible in terms of code. However, it changes some directory names in `download`:
(1) `Orchset` becomes `ORCHSET`
(2) `Salami` becomes `SALAMI`
(3) `Groove MIDI` becomes `Groove-MIDI`

(1) and (2) is for consistency of capitalization with the official name
(3) is to avoid spaces in file paths

I wrote new tests in `test_track()` to ensure that this consistency is enforced in the future. The `name` must:
* only contains alphanumeric characters, hyphens, underscores, or spaces
* begin with a letter (this is important for Python module names)
* be at least three characters long
* end with an alphanumeric character

Obtaining the `DATASET_DIR` from the `name` is deterministic
Obtaining the `module_name` from the `DATASET_DIR` is deterministic


rabitt:
@lostanlen can we wait on this - I don't have a ton of time and this is not a prioritized issue.

Issue #224, opened 2020-04-08
What information can be abstracted to the track.Track object?


rabitt:
The generic `mirdata.track.Track` object provides a simple abstraction in order to automate the `__repr__` method which we used to write manually for each dataset's module.

In #219 @lostanlen experimented with what other information we could abstract in order to simplify each module's Track object. He proposed to have the following live as attributes/properties/methods of the new Track object:
* `audio` which loads an attribute in the track index called audio
* `duration` (pulling from either the metadata or from audio)
* a `from_jams` method which instantiates mirdata objects from a jams object
* a `load_metadata` method, which would live at the track-level rather than the module level
* the `to_jams` method by programatically parsing the track attributes
* a `validate` method, which would live at the track-level rather than the module level
* the entire init function, by instantiating the object with a track_metadata dictionary

Any thoughts about each of these abstractions?

(@lostanlen correct any of this if I got the details wrong)

lostanlen:
Thank you for raising this. Yes, all of this is correct. Here are some details:

* `audio` is really an easy one. It relies on the fact that mirdata indexes have keys to describe them. So, if we encourage index providers to denote their audio files by the key "audio" (and this doesn't sound like a eccentric thing to ask), then this audio is immediately available through the cached property `track.audio`, which is implemented in the parent class. So it ends up being less code for the contributors, because they don't need to implement "audio" (nor `load_audio`). If the Track in question has no audio, well: `raise AttributeError`. Then, for fancier datasets which have more than one kind of audio (e.g. GuitarSet), the dataset providers may add other properties on top: `audio_hex`, `audio_mic`, etc. But there can only be one `audio` property; and by convention, we assume that it's where the file indexed as `audio` lives.

* `duration`, on the contrary, is the hardest one. Is it mutable or immutable? It is a property or an attribute? Should we cache it? Suppose that we have an annotation on disk but we haven't downloaded the audio: should we raise an `AttributeError` until we can get the duration from the audio, or have a guesstimate of the duration via the annotation (e.g. time of the last beat?) These are difficult design questions, and i don't have a set-in-stone answer. The reason why i wrote a hacky version of `duration` is that i needed it for `to_jams` to validate.

* `a load_metadata method, which would live at the track-level rather than the module level``
[...]

PR #235, opened 2020-04-09
[RFC] move midi property to Track super class

lostanlen:
just an idea; let me know what you think
in the spirit of #224, i propose to move Groove MIDI's `midi` property to the Track parent.
This is backwards compatible, has a negative LOC net count, and passes tests (at least on my machine).

The idea is that, if an index has a `track_path` indexed by the word `midi`, then we use `pretty_midi` to load it automatically. It generalizes the Groove MIDI pattern to potential future MIDI datasets.

Are you OK with this abstraction or is that already to much to assume about future datasets?

### DIFF
--- a/mirdata/groove_midi.py
+++ b/mirdata/groove_midi.py
@@ -286,8 +286,6 @@ class Track(track.Track):
         self.midi_filename = self._track_metadata["midi_filename"]
         self.audio_filename = self._track_metadata["audio_filename"]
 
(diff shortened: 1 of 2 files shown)


lostanlen:
huh, i was wrong about that. tests are failing for `Track.__repr__` :(

lostanlen:
nevermind. the Track class is not ready yet for things like this. it's not hogh priority anyway

PR #238, opened 2020-04-10
Add local exhaustive dataset tests

### LINKED ISSUE #234: Local tests to load all tracks as part of new_loader review
As part of review of a `new_loaders`, we should run local tests to make sure all tracks load annotations and metadata as expected. That will help catch bugs before merging into master.

Edit [rabitt]: Let's add a test to `test/test_loaders.py` which runs download + load on the full dataset, and decorate it with a `@pytest.mark.skipif(local=False)`, setting `local=True` with a pytest parameter:
http://doc.pytest.org/en/latest/example/parametrize.html

And since in practice we'll be testing on particular new dataset at a time, we should pass a second optional parameter to pytest which specifies which dataset we should run the long tests on.

rabitt:
fixes #234

What changed:
* I've added `tests/test_full_dataset.py`, which downloads, validates and loads a dataset, and for each track object, calls all properties to make sure they are callable.
* we can now run pytest with an optional `--local` argument (in `tests/conftest.py`) which skips any tests that are built for the remote tests and always fail
* I've decorated a few of the existing tests to either skip when run remotely or skip when run locally, when it makes sense
* Updated documentation in `CONTRIBUTING.md` to explain how to use the `--local` flag and how to run the new full test.

Once this is merged, we should (ugh...) run this for every dataset and make sure there are no secret outstanding bugs.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+def pytest_addoption(parser):
+    parser.addoption("--local", action="store_true", default=False, help="run local tests")
+    parser.addoption("--dataset", type=str, default="", help="dataset to test locally")
+    parser.addoption("--skip-download", action="store_true", default=False, help="skip download step")
+
+
+# @pytest.fixture(scope='session')
+# def local(request):
+#     return request.config.getoption('--local')
+
+@pytest.fixture(scope='session')
+def skip_local(request):
+    if request.config.getoption('--local'):
+        pytest.skip()
+
+
+@pytest.fixture(scope='session')
+def skip_remote(request):
+    if not request.config.getoption('--local'):
+        pytest.skip()
+
+
+@pytest.fixture(scope='session')
+def test_dataset(request):
+    return request.config.getoption('--dataset')
+
+
+@pytest.fixture(scope='session')
+def skip_download(request):
+    return request.config.getoption('--skip-download')
--- a/tests/test_full_dataset.py
+++ b/tests/test_full_dataset.py
@@ -25,11 +25,6 @@ def data_home_dir(dataset):
 def test_download(skip_remote, dataset, data_home_dir, test_dataset, skip_download):
     assert test_dataset in mirdata.__all__
 
-    # dataset = importlib.import_module("mirdata.{}".format(test_dataset))
-
-    # # new directory to test download
-    # data_home_dir = os.path.join('tests/resources/mir_datasets_full', dataset.DATASET_DIR)
-
     # download the dataset
     if not skip_download:
         dataset.download(data_home=data_home_dir)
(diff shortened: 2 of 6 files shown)


tests/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+    parser.addoption("--skip-download", action="store_true", default=False, help="skip download step")
+
+
+# @pytest.fixture(scope='session')
+# def local(request):
+#     return request.config.getoption('--local')
+
+@pytest.fixture(scope='session')
+def skip_local(request):
+    if request.config.getoption('--local'):
```
magdalenafuentes, line 16: Maybe this is right but is just a bit unintuitive for me: you skip the local tests if you get the `--local` flag and skip the remote if you don't? or I completely misunderstood?

tests/test_full_dataset.py
```diff
@@ -0,0 +1,76 @@
...
+
+# This is magically skipped by the the remote fixture `skip_remote` in conftest.py
+# when tests are run without the --local flag
+def test_download(skip_remote, dataset, data_home_dir, test_dataset, skip_download):
+    assert test_dataset in mirdata.__all__
+
+    # dataset = importlib.import_module("mirdata.{}".format(test_dataset))
+
+    # # new directory to test download
+    # data_home_dir = os.path.join('tests/resources/mir_datasets_full', dataset.DATASET_DIR)
```
magdalenafuentes, line 31: should remove these maybe?

tests/test_full_dataset.py
```diff
@@ -0,0 +1,76 @@
...
+    track_ids = dataset.track_ids()
+    assert set(track_ids) == set(all_data.keys())
+
+    #test that all attributes and properties can be called
+    for track_id in track_ids:
+        track = all_data[track_id]
+        track_data = get_attributes_and_properties(track)
+
+        for attr in track_data['attributes']:
+            ret = getattr(track, attr)
```
magdalenafuentes, line 67: question: would this break if this returns an `AnnotationType` with nothing inside (i.e. `None`)? That's what you get in bugs such as the Groove MIDI folder structure

tests/conftest.py
rabitt: It's confusingly written. The `--local` just means that it's being run locally. Then there are two fixtures `skip_local` and `skip_remote` which can be passed to different tests. When a test is passed `skip_local`, it will only skip it when `--local` is passed. When a test is passed `skip_remote`, it will only skip it when `--local` is not passed.

tests/test_full_dataset.py
rabitt: It shouldn't break, it should just return None. The issue you're talking about should (hopefully) be fixed by #240 , so that if a file is actually missing, an error will be raised, and we'll catch it in this test.

magdalenafuentes: Then this is good to go!

rabitt: @magdalenafuentes feel free to merge at your convenience!

PR #239, opened 2020-04-10
Fix groove_midi download structure

### LINKED ISSUE #236: groove_midi.load() fails
The folder structure where `Groove MIDI` data is downloaded it's not right, now it's 
 `mir_datasets/Groove MIDI/Groove MIDI` but the loader expects the files to be one level up.

rabitt:
Fixes #236 

What changed: 
* The groove_midi download function downloads the zip file which creates a subfolder called `groove`. We copy its contents to the parent directory and delete it.
* I added a test with a tiny zipfile having the same structure as the real one which tests if the change works
* The value of DATASET_DIR changed from `Groove MIDI` (which has a space, so it can behave badly when part of a path) to `Groove-MIDI`.

If we wait for #238 to be merged, we can test this in the real case too.


mirdata/groove_midi.py
```diff
@@ -43,19 +43,21 @@
...
+import shutil
 
 import mirdata.download_utils as download_utils
 import mirdata.jams_utils as jams_utils
 import mirdata.track as track
 import mirdata.utils as utils
 
 
-DATASET_DIR = 'Groove MIDI'
+DATASET_DIR = 'Groove-MIDI'
```
lostanlen, line 60: i love these space-free paths ❤️

mirdata/groove_midi.py
```diff
@@ -378,7 +380,15 @@ def download(data_home=None):
         data_home, zip_downloads=[AUDIO_MIDI_REMOTE], cleanup=True
     )
 
-    os.rename(os.path.join(data_home, 'groove'), os.path.join(data_home, 'Groove MIDI'))
+    # files get downloaded to a folder called groove - move everything up a level
```
lostanlen, line 383: thanks. this is a bit slower but indeed the safest option, considering that we can't guarantee that we have write access to `os.path.split(data_home)[0]`

tests/test_groove_midi.py
```diff
@@ -78,3 +80,32 @@ def test_load_metadata():
...
     assert metadata_none is None
+
+
+def test_download(httpserver):
+    data_home = 'tests/resources/mir_datasets/Groove-MIDI_download'
+    if os.path.exists(data_home):
+        shutil.rmtree(data_home)
+
+    httpserver.serve_content(
+        open('tests/resources/download/groove-v1-0.0.zip', 'rb').read()
```
lostanlen, line 91: amazing!!

PR #240, opened 2020-04-10
module.load_* functions raise an IOError when file does not exist

### LINKED ISSUE #222: `module.load_*` should raise `IOError`, not return `None`
raised by @rabitt in a meeting today

Many of the module-specific loaders, e.g. `orchset.load_melody`, return `None` if the argument does not exist. This can be confusing in the sense that it may suggest that the file exists but that there is no annotation for this file, or that the annotation is not applicable (non-melodic music, following my example above).

She now recommends raising an `IOError`. 

This is a breaking change so i wouldn't recommend it for v0.2
But we can deprecate it in v0.2 and change the behavior in v0.3

@rabitt what should we do for people providing new loaders? Should we tell them to work with the old or the new convention? 
This affects `new_loader.md` (excerpt below):

```
# -- Write any necessary loader functions for loading the dataset's data
def load_annotation(annotation_path):
    if not os.path.exists(annotation_path):
        return None
```

rabitt:
Fixes #222 

What changed:
* all `dataset.load_*` functions now raise an IOError if passed a filepath which does not exist
* Some exceptional `dataset.load_*` function only returns `None` when passed `filepath=None`, for datasets where there are missing files for some tracks (e.g. salami in `load_sections`, groove_midi in `load_audio`)
* `tests/test_loaders.py` has an automated test for any dataset's load function to catch that it raises an IOError
* removes now redundant tests and updates tests that were checking that nonexistent filepaths returned `None`


CONTRIBUTING.md
```diff
@@ -255,6 +255,8 @@ def load_audio(audio_path):
     # -- for example, the code below. This should be dataset specific!
     # -- By default we load to mono
     # -- change this if it doesn't make sense for your dataset.
+    if not os.path.exists(audio_path):
+        raise IOError("audio_path {} does not exist".format(audio_path))
```
lostanlen, line 259: what if the file exists, but we don't have read permission?

lostanlen: also, what if it exists, has read permissions, but has the wrong checksum?

rabitt: > what if the file exists, but we don't have read permission?

We won't catch it here, but the function will still error later when we open the file, just with a less friendly error message. I think it's an unusual enough case that I'm OK with not being 100% exhaustive of all possible cases. 

> also, what if it exists, has read permissions, but has the wrong checksum?

We used to validate on load, but it was _very_ slow and made the library much less usable. @magdalenafuentes and I made the sad (but in our opinion) necessary decision to leave validation up to the user. This is what `dataset.validate()` is for!

mirdata/guitarset.py
```diff
@@ -215,7 +214,7 @@ def notes(self):
         notes = {}
         # iterate over 6 strings
         for i in range(6):
-            notes[_GUITAR_STRINGS[i]] = _load_note_ann(self.jams_path, i)
+            notes[_GUITAR_STRINGS[i]] = load_note_ann(self.jams_path, i)
```
magdalenafuentes, line 217: good catch

PR #247, opened 2020-04-16
fixes #226

### LINKED ISSUE #226: Jams object completeness
For ease of contribution, we've implemented a `jams_utils.jams_converter` function which writes jams objects under the hood for each Track object's `.to_jams()` method. This has two benefits:
* contributors do not need to be familiar with the jams api
* each dataset module's code is drastically simplified.

However, this limits the completeness of the resulting jams object - namely, we omit much of the metadata that could be included. We now enforce that the jams objects returned by `Track.to_jams()` passes validation, so they are minimally usable, but they are not ideal.

Do we have any ideas for how to improve the jams object completeness without hurting the ease of contribution? Do we need fully complete jams files for most use cases (e.g. usage with `muda`)

rabitt:
Fixes #226

What changed:
* the jams_converter now takes `audio_path` as an argument which is used to fill the duration value of the resulting jams object
  * this removes the need for computing duration in the individual `to_jams()` methods
  * we can catch the OSError within the `jams_converter` function itself
* updated tests
* updated contributing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -208,9 +208,9 @@ class Track(track.Track):
             self._data_home, self._track_paths['annotation'][0])
 
         # -- if the user doesn't have a metadata file, load None
-        metadata = DATA.metadata(data_home)
-        if metadata is not None and track_id in metadata:
-            self.some_metadata = metadata[track_id]['some_metadata']
+        self._metadata = DATA.metadata(data_home)
+        if self._metadata is not None and track_id in self._metadata:
+            self.some_metadata = self._metadata[track_id]['some_metadata']
         else:
             self.some_metadata = None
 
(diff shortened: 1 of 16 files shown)


CONTRIBUTING.md
```diff
@@ -237,10 +237,12 @@ class Track(track.Track):
     # -- object with the annotations.
     def to_jams(self):
         """Jams: the track's data in jams format"""
-        return jams_utils.jams_convertrer(
+        return jams_utils.jams_converter(
+            audio_path=self.audio_path,
             annotation_data=[(self.annotation, None)],
-            metadata=metadata},
+            metadata=self._metadata,
```
magdalenafuentes, line 243: nice one

mirdata/jams_utils.py
```diff
@@ -62,14 +71,39 @@ def jams_converter(
...
+    if audio_path is not None:
+        if os.path.exists(audio_path):
+            duration = librosa.get_duration(filename=audio_path)
+        else:
+            raise OSError(
+                'jams conversion failed because the audio file '
+                + 'for this track cannot be found, and it is required'
+                + 'to compute duration.'
+            )
+
```
magdalenafuentes, line 85: :star2:

lostanlen: does line 78 potetntialltly raises errors? if so, better to document them

PR #285, opened 2020-10-10
Adding loader for Mridangam Stroke Dataset

genisplaja:
### Adding loader for Mridangam Stroke Dataset

 Please use the following title: "Adding loader for MyDATASET". If your pull request is work in progress, change your title to "[WIP] Adding loader for MyDATASET" to avoid reviews while the loader is not ready.

### Description

Please include the following information at the top level docstring for the dataset's module mydataset.py:

- [x] Describe annotations included in the dataset
- [x] Indicate the size of the datasets (e.g. number files and duration, hours)
- [x] Mention the origin of the dataset (e.g. creator, institution)
- [x] Describe the type of music included in the dataset
- [x] Indicate any relevant papers related to the dataset
- [x] Include a description about how the data can be accessed and the license it uses (if applicable)

#### Dataset loaders checklist:

- [x] Create a script in `scripts/`, e.g. `make_my_dataset_index.py`, which generates an index file.
- [x] Run the script on the canonical version of the dataset and save the index in `mirdata/indexes/` e.g. `my_dataset_index.json`.
- [x] Create a module in mirdata, e.g. `mirdata/my_dataset.py`
- [x] Create tests for your loader in `tests/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/datasets.rst`
- [x] Add the module to `mirdata/__init__.py`
- [x] Add the module to the list in the `README.md` file, section `Currently supported datasets`

This dataset is **fully downloadable** .

#### Please-do-not-edit flag
To reduce friction, we will make commits on top of contributor's pull requests by default unless they use the `please-do-not-edit` flag. If you don't want this to happen don't forget to add the flag when you start your pull request.

PR #288, opened 2020-10-16
Adding loader for IRMAS Dataset

genisplaja:
### Description:

- [x] Create tests for your loader in `tests/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/datasets.rst`
- [x] Add the module to `mirdata/__init__.py`
- [x] Add the module to the list in the `README.md` file, section `Currently supported datasets`

#### Please-do-not-edit flag
To reduce friction, we will make commits on top of contributor's pull requests by default unless they use the `please-do-not-edit` flag. If you don't want this to happen don't forget to add the flag when you start your pull request.


nkundiushuti:
> # [Codecov](https://codecov.io/gh/mir-dataset-loaders/mirdata/pull/288?src=pr&el=h1) Report
> > Merging [#288](https://codecov.io/gh/mir-dataset-loaders/mirdata/pull/288?src=pr&el=desc) into [master](https://codecov.io/gh/mir-dataset-loaders/mirdata/commit/46ce5692e07e77cd2e1ecc2aecff80c36f7cd279?el=desc) will **decrease** coverage by `0.92%`.
> > The diff coverage is `78.63%`.
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master     #288      +/-   ##
> ==========================================
> - Coverage   98.97%   98.04%   -0.93%     
> ==========================================
>   Files          21       23       +2     
>   Lines        2246     2457     +211     
> ==========================================
> + Hits         2223     2409     +186     
> - Misses         23       48      +25     
> ```

@genisplaja , you need to put tests for all the methods you declare in the dataset

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
...
+                                if file.endswith('.wav'):
+                                    test_index = str(count)
+                                    irmas_dict[test_index] = {
+                                        'genre': None,
+                                        'drum': None,
+                                        'train': False,
+                                    }
+                                    count += 1
+
+    irmas_dict["data_home"] = data_home
```
magdalenafuentes, line 179: I think this code can be simplified but I want to make sure I understand what you're doing first. I guess the metadata is contained in the `wavfile` name, right? So you use split to get the different components?

If I got that right, I think you can look over the track ids, and simply change the dict as you go:

``` 
for track_id in INDEX.keys():
        irmas_dict[track_id]={
                              'genre': split(INDEX[track_id].audio_path, bla bla). ,
                              'drum': [True if 'dru' in INDEX[track_id].audio_path else False][0],
                              'train': [True if 'Train' in INDEX[track_id].audio_path else False][0]
                              }
```
Or something like that. Does this make sense?

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
...
+
+        self.annotation_path = os.path.join(
+            self._data_home, self._track_paths['annotation'][0]
+        )
+        self.genre = self._track_metadata['genre']
+        self.drum = self._track_metadata['drum']
+        self.train = self._track_metadata['train']
+
+    @utils.cached_property
+    def predominant_instrument(self):
```
magdalenafuentes, line 232: If there's only one instrument per recording, I'll name this `instrument` to be consistent with other datasets

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
...
+    @property
+    def audio(self):
+        """(np.ndarray, float): audio signal, sample rate"""
+        return load_audio(self.audio_path)
+
+    def to_jams(self):
+        """Jams: the track's data in jams format"""
+        return jams_utils.jams_converter(
+            audio_path=self.audio_path,
+            event_data=[(self.predominant_instrument, 'predominant_instrument')],
```
magdalenafuentes, line 245: I'm not too familiar with this dataset, but I'm curious about why this annotations are `eventdata`?

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
...
+        with open(annotation_path, 'r') as fopen:
+            pred_inst_file = fopen.readlines()
+            pred_inst = []
+            for inst_ in pred_inst_file:
+                inst_code = inst_[:3]
+                pred_inst.append(inst_code)
+
+        return utils.EventData(
+            np.array([0]), np.array([duration]), np.array([tuple(pred_inst)])
+        )
```
magdalenafuentes, line 383: I know we went back and forth with this and sorry if it is confusing! Finally we converged that if the annotation doesn't change over time, we will keep it same way we keep metadata. So if you had only one instrument, you would return a string, and in this case that you could get multiple, I guess you would have a list of strings? Does this make sense?

mirdata/irmas.py
magdalenafuentes: See comment below

magdalenafuentes (changes requested):

mirdata/irmas.py
nkundiushuti: I also think you should look at the index rather than at the files that are on the hard disk

genisplaja: Yes, currently fixing it. I haven't though that I could use the json index here! However, I'll need to add some conditions as the files have to be split from different points because the 'dru' (has drum) or 'nod' (has not drum) annotation is not in every file.

mirdata/irmas.py
genisplaja: In fact, the purpose of this dataset is to detect the predominant instrument(s) in a track. In many tracks, most of them (especially in the testing set), the tracks have many instruments and the idea is to detect the most prominent one. So should I leave it this way or should I call name it `instrument` and then specify in the documentation that it is referring to the predominant one?

mirdata/irmas.py
genisplaja: Ok! It does make sense, parsing the data will turn more straightforward. So should I add the instrument annotation to _track_metadata? I will implement int this way and see how it looks like!

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
genisplaja: Fixed!

docs/source/datasets.rst
```diff
@@ -214,6 +217,11 @@ of annotation is implicitly available when Vocal F0_ or Vocal Notes_ annotations
 .. _gtzan_genre: https://mirdata.readthedocs.io/en/latest/source/mirdata.html#module-mirdata.gtzan_genre
 .. _guitarset: https://mirdata.readthedocs.io/en/latest/source/mirdata.html#module-mirdata.guitarset
 .. _ikala: https://mirdata.readthedocs.io/en/latest/source/mirdata.html#module-mirdata.ikala
+<<<<<<< HEAD
+.. _irmas: https://mirdata.readthedocs.io/en/latest/source/mirdata.html#module-mirdata.irmas
+=======
+.. _irmas: https://www.upf.edu/web/mtg/irmas
+>>>>>>> 8d6f3436fb4a7583ed9aab596b7036ab1158d00b
```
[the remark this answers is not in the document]
genisplaja: Fixed!

mirdata/irmas.py
genisplaja: So, the issue here is the following: if I want to return the instrument annotation as a String, there is no argument in jams_converter that accepts it, apart from adding it to the metadata together with 'genre' and 'drum'. 
However, although I can retrieve the predominant instrument annotation from the training samples filename in the index, I need to access all the .txt files of the testing samples annotations to retrieve the data. For that reason, when trying to add the predominant instrument to the metadata, the pytest fails because it requires the whole dataset to pass. That is why, storing `predominant_instrument` as `EventData` was working well, because I could use the function to load the instrument annotation in the `@utils.cached_property`. If I try to load the instrument annotation directly from the Track class in order to add it to the metadata, the pytests are failing because I need the whole dataset locally (or at least that what I understand). Am I correct? Did I explain myself?

mirdata/irmas.py
genisplaja: Solved! See [commit 45a71fd](https://github.com/mir-dataset-loaders/mirdata/pull/288/commits/45a71fd477006900ac2701e59c139968d3ab0e82).

mirdata/irmas.py
```diff
@@ -0,0 +1,406 @@
...
+It was used for the evaluation in the following article:
+Bosch, J. J., Janer, J., Fuhrmann, F., & Herrera, P. “A Comparison of Sound Segregation Techniques for
+Predominant Instrument Recognition in Musical Audio Signals”, in Proc. ISMIR (pp. 559-564), 2012.
+
+IRMAS is intended to be used for training and testing methods for the automatic recognition of predominant
+instruments in musical audio. The instruments considered are: cello, clarinet, flute, acoustic guitar,
+electric guitar, organ, piano, saxophone, trumpet, violin, and human singing voice.
+This dataset is derived from the one compiled by Ferdinand Fuhrmann in his PhD thesis, with the difference
+that we provide audio data in stereo format, the annotations in the testing dataset are limited to specific
+pitched instruments, and there is a different amount and lenght of excerpts.
```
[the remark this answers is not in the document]
genisplaja: Oh! This is just to remark that this dataset is derived from the one that Ferdinand Fuhrmann compiled for his PhD thesis. This last sentence is just stating the differences between the IRMAS dataset that I am integrating and the original dataset complied by F. Fuhrmann, so at the end of the day, this information is far from being crucial to understand and use the dataset correctly (although is very important to know where the dataset comes from and who are the creators and compilers!). In fact, F. Fuhrmann and the rest of researchers involved are correctly cited in the code. 
Maybe just writing the last sentence as following "and there is a different amount and length of excerpts from the original dataset" would help? I don't think I can access to the original dataset, but I guess that it was bigger and excerpts were longer. I'll ask the department for more info about this!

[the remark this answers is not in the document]
magdalenafuentes: Yes that would help! And thanks for the explanation :)

mirdata/irmas.py
magdalenafuentes: Mmmm... I see. So you have many instruments per track, but you're interested in finding the predominant one, right? Would it make sense to have **both** `predominant_instruments` **and** `instruments`? In the first one, you have the target instrument, and in the second, you have all instruments in the track (for e.g. instrument recognition). Just an idea, don't know if it makes sense for this dataset, maybe @nkundiushuti has comments on this?

genisplaja: The problem here is that there is no annotation of the complete instrument line-up of the tracks. For both training and testing samples, we just have the predominant instrument, without having any information about the rest of the  background instruments.

mirdata/irmas.py
```diff
@@ -112,69 +112,31 @@
...
+                        'genre': genre,
+                        'drum': None,
+                        'train': True
+                    }
+            else:
+                irmas_dict[track_id] = {
+                    'genre': None,
+                    'drum': None,
+                    'train': False
+                }
```
magdalenafuentes, line 139: This is looking much better! I think I understand better the structure as well.

Ideally we can use `DATA = utils.LargeData("tinysol_index.json", _load_metadata)` which takes care of loading both the index and the metadata (and they are accessed as `DATA.index` and `DATA.metadata`, see [here for an example](https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/tinysol.py)). However this dataset seems tricky since you have to parse the wav files. Now that the code is simplified let's see if we can change things to use the `LargeData` object.  Can you explain what's the directory structure a bit?

mirdata/irmas.py
magdalenafuentes: Mmmm... this is because of using the index in the `_load_metadata` right? I think it might be solved with my other comment in the implementation of that function

mirdata/irmas.py
magdalenafuentes: Ooh I see, thanks. It seems this case is different from the datasets we already have with instruments (`tinysol`, `medley_db_solos` and `medley_db_pitch`). @rabitt should we go for `instrument` to be compatible with those even though there are multiple instruments in the mix, but we "care" only about one?

mirdata/irmas.py
genisplaja: I would say that this is because of parsing annotation files (in this case the .txt files) in the Track class itself. I mean, running the `load_pred_inst` function in the Track class to parse the intrument annotation and insert it into `_track_metadata` is not passing the tests, especially the following one: `pyest tests/irmas.py --local`. I believe that doing it inside `_load_metadata` would fail as well, but I haven't tried tbh. I will take a deep look into the other comment about `LargeData` to try to solve this! I am not completely sure about that assumptions... at the end you are way more familiar than me with mirdata...!!!

genisplaja: But I am saying that because when running pytests on IRMAS dataset, if I parse information in the Track class, I get an error because the loader is looking for the whole set of annotations. Note that _predominant instrument_ is the only annotation in this dataset that is written in a single .txt file for each testing track. How could we add it into the metadata?

mirdata/irmas.py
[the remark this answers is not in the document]
genisplaja: Cool! I'll commit the update now.

mirdata/irmas.py
genisplaja: So you would have a general IRMAS folder. Into the IRMAS folder there are 4 folders:

- IRMAS-TrainingData
Inside this folder you can find several folders, each of them referring to a different instrument. Exemples: cla/, flu/, etc. The samples inside flu/ will be tracks were a flute is the predominant instrument. This track have all the annotations in the filename: predominant instrument, genre and drum presence (sometimes), plus an unique ID.


- IRMAS-TestingData-Part1/Part1 (inside the TestingData folders there is a README file with the information written in the documentation and then a folder named PartX with the files)
- IRMAS-TestingData-Part2/Part2
- IRMAS-TestingData-Part2/Part3
Inside IRMAS-TestingData-PartX/PartX you can find the testing samples belonging to that part of the testing data. They are random song excerpts from different musical styles and origins. Every .wav track is accompanied with a .txt file (that has the same name) with the annotation of the predominant instrument(s). Example: Highway_to_hell.wav and Highway_to_hell.txt, and inside the .txt file you would find "voi /n gel" (voi refers to voice and gel to electric guitar).

Here you can find the [Zenodo link to the dataset](https://zenodo.org/record/1290750#.X47x51ngq8U) in case my explanation is confusing or not enough clear:)

mirdata/irmas.py
genisplaja: Could we insert the instrument data as "tag open"? That is what we are doing for the [Mridangam Stroke Dataset](https://github.com/mir-dataset-loaders/mirdata/pull/285/) (that's the link of the Mridangam Stroke PR). This dataset has algo an annotation that lasts the same for the whole track!

genisplaja:
I have just fixed the whole thing a bit... following your indications @magdalenafuentes. What do you think about the fixes? 
- I changed `predominant_instrument` to `instrument`. At the documentation is well explained that it refers to the predominant one.
- I stored `instrument` into the metada together with the additional annotations, as a list of Strings. 
- I am using `DATA.metada` to get the annotations, I hope that is what you were asking for!
I am requesting you to review it again:) Thank you very much!!!!!!!!!!

magdalenafuentes:
Hey @genisplaja sorry for the delay, rushing a deadline. I'm gonna take a look now!

mirdata/irmas.py
magdalenafuentes: @rabitt, apparently to get the metadata for this dataset as we usually do we should parse a bunch of files down in several folders as @genisplaja was doing, but is a lot of code and walk through directories. Instead we're parsing the index, but that goes against what we usually do that is to use the `LargeData` object: in this case we're parsing the index in `_load_metadata` **and** using the `DATA` object. What are your thoughts on this now that you're doing the `Dataset` object?

magdalenafuentes (commented):
It's looking really good!

magdalenafuentes:
Let's wait @rabitt comment on the `_load_metadata` otherwise looking good!

genisplaja:
Thanks @magdalenafuentes for your feedback!! No worries about the delay, it's alright. Please take your time:) Looking forward to recieve more thoughts about the actual implementation!

tests/test_giantsteps_key.py
```diff
@@ -13,11 +13,11 @@ def test_track():
 
     expected_attributes = {
         'audio_path': 'tests/resources/mir_datasets/GiantSteps_key/audio/10089 Jason Sparks - Close My Eyes feat. J. '
-                      'Little (Original Mix).mp3',
```
nkundiushuti, line 16: this file should not be in this PR

mirdata/irmas.py
rabitt: I think this dataset doesn't actually need a `_load_metadata` function at all. Lines 118-138 can be moved almost exactly into the `__init__` of the `Track` object, replacing lines 182-193! since everything just depends on the index and not on any external metadata file, you have all the information you need directly inside the Track object (via `self._track_paths`). It will make the init a bit more verbose, but its ok!

genisplaja: Ok! Fixed now! Sorry for the delay.

mirdata/irmas.py
```diff
@@ -0,0 +1,349 @@
...
+    if data_home is None:
+        data_home = utils.get_default_dataset_path(DATASET_DIR)
+
+    data = {}
+    for key in DATA.index.keys():
+        data[key] = Track(key, data_home=data_home)
+    return data
+
+
+def load_pred_inst(audio_path, annotation_path, train):
```
nkundiushuti, line 296: maybe I missed some discussion before, but isn't this supposed to be the folder where the audio file corresponding to current Track is stored? 
this can be obtained easily with 
`os.path.basename(os.path.dirname(audio_path))`

nkundiushuti (changes requested):
remove some unrelated code 
validate genre using a pre-defined list of all genres in dataset
check to see if load_pred_inst can be done using os.path.basename

nkundiushuti:
I have two errors while testing the full dataset. Just after download:

E       AssertionError: assert {'1005': ['te...-6.txt'], ...} == {}
E         Left contains 36 more items:
E         {'1005': ['tests/resources/mir_datasets_full/IRMAS/IRMAS-TestingData-Part3/Part3/10 '
E                   '- Johann Sebastian Bach - English Suite No. 3 in G major, BWV 808 '
E                   '(Andràs Schiff) (1)-1.wav',
E                   'tests/resources/mir_datasets_full/IRMAS/IRMAS-TestingData-Part3/Part3/10 '
E                   '- Johann Sebastian Bach - English Suite No. 3 in G major, BWV 808 '
E                   '(Andràs Schiff) (1)-1.txt'],...
E         
E         ...Full output truncated (188 lines hidden), use '-vv' to show

tests/test_full_dataset.py:57: AssertionError

E       FileNotFoundError: [Errno 2] No such file or directory: 'tests/resources/mir_datasets_full/IRMAS/IRMAS-TestingData-Part1/Part1/01 - Canto das três raças-9.txt'

mirdata/irmas.py:370: FileNotFoundError

Do I need to do something manually for the dataset to work after downloading?

mirdata/irmas.py
rabitt: Quick comment here - the load_* functions should be as 'generic' as possible. I don't know this dataset well, but I don't think it's good to be passing `audio_file` just to parse the path. Is it the case that test tracks have annotation files, while training tracks don't, and the label is inferred from the filename? What is the difference between the test track file names and the test track annotations?

It looks like during training, you use the name of the file to parse the predominant instrument - to me this should just be in the track metadata, not part of this load function. During testing, it loads an annotation file and this part seems perfectly fine. If there are no annotation files for some tracks, we have other loaders that return e.g. `None` when the file does not exist (see `salami.py` for some examples)

My proposal is this function should just have:
```
def load_pred_inst(annotation_path):
    with open(annotation_path):
        ...
    # plus some possible check for if annotation_path is None
```

Then, in `Track.__init__` you could create an attribute `self._predominant_instrument=audio_path.split('[')[1].split(']')[0]` 

In `self.instrument`, you can return `self._predomiant_instrument` if it's set, and load from the annotation otherwise.

genisplaja: Ok, done!! The only issue here is that now Track has `instrument` with the list of strings as the "official" annotation, and `predominant_instrument` which is the attribute I added to `Track.__init__`. Is this Ok @rabitt? Thanks for the feedback:)

nkundiushuti:
this is currently blocked by #306

genisplaja:
I am closing this PR and opening a new one with the [IRMAS branch I just created](https://github.com/mir-dataset-loaders/mirdata/tree/genis/irmas), to adapt better to the new Dataset object and pull easily new changes to come.
The PR status is the same as the one for this branch! It seems that it is ready to merge but waiting for @nkundiushuti's PR about Windows encoded files fix to be previously merged.
Thanks everyone for the reviews and comments! Feel free to add more comments or reviews to the new PR if necessary.

PR #295, opened 2020-10-20
MultiTrack class

### LINKED ISSUE #276: Multitrack Datasets
How can we best support multitrack datasets? The current solution is to have each 'Track' be a multitrack with a ton of attributes, but it's clunky and difficult to take full advantage of the multitracks themselves. We're also loosely tying a `Track` to something that can be mapped to a `jams` file/object, and jams isnt built for multiple audio files.

My current thinking is we could index at the stem level, and support a new base class `MultiTrack` which would group multiple `Track` objects. The grouping could be stored as part of the index, and any mixture-level annotations could itself be stored as a `Track` object, and be a part of the multitrack object.

Thoughts?

cc @nkundiushuti  - we could test this out with Phenicx-Anechoic #270

rabitt:
Resolves #276


mirdata/track.py
```diff
@@ -33,11 +33,78 @@ def __repr__(self):
...
+            track_keys (list): list of track keys to mix together
+            weights (list or None): list of positive scalars to be used in the average
+        
+        Returns:
+            target (np.ndarray): mixture audio with shape (n_samples, n_channels)
+        """
+        signals = [
+            getattr(self.tracks[k], self.track_audio_attribute)() for k in track_keys
+        ]
+        return np.average(signals, axis=0, weights=weights)
```
nkundiushuti, line 78: sometimes we don't want to average the signals and take the sum (multichannel)
in this case maybe we can add a parameter average=True 
`if average:
    return np.average(signals, axis=0, weights=weights)
else:
    return np.sum(signals, axis=0, weights=weights)`

nkundiushuti: also this would give an error if the signals have different lengths

mirdata/track.py
```diff
@@ -33,11 +33,78 @@ def __repr__(self):
...
             repr_str += "  {}: {},\n".format(prop, val_type_str)
 
         repr_str += ")"
         return repr_str
 
     def to_jams(self):
         raise NotImplementedError
+
+
+class MultiTrack(Track):
```
nkundiushuti, line 46: just a quick question: 
I am interested in score-informed source separation. in this case, each track and target should have annotations (scores).
so in my case, I should extend this the Track and MultiTrack objects with score-related methods, similarly to other mirdatasets loaders which deal with their particular annotations?

mirdata/track.py
rabitt: OK, will update!

+1 for catching arrays with different lengths

mirdata/track.py
rabitt: yes exactly!

nkundiushuti:
> @nkundiushuti should be ready to go, let me know if this will work for #270 or if there's something missing

thanks! I will check this!

nkundiushuti (approved):
this looks excellent. I could do a simpler implementation of Phenicx-Anechoic thanks to this PR :) thanks, Rachel!

rabitt:
Merging!

Issue #293, opened 2020-10-19
Generalize copy-paste dataset functions to utils?


rabitt:
`dataset.validate()`, `dataset.load()` and `dataset.track_ids()` are identical in every loader. How can we generalize this?

Option 1 - lambdas
-------------------
we define in e.g. `utils` lambda functions, which get instantiated in each dataset, e.g.
`track_ids = utils.track_ids(DATA.index)`
where `utils.track_ids` is itself a lambda function

pros: same API, less code to copy paste
cons: still copy pasting code, not very nice coding style

Option 2 - reverse the api
-------------------------
`utils.validate('orchset')` rather than `orchset.validate()`

pros: no code copy-pasting, better code style
cons: changes the API, makes some inconsistency - some things are dataset-major (e.g. orchset.Track) some are function-major (utils.validate('orchset'))

Option 3 - split it all up
-----------------------
*move validate to a top level (utils.validate('orchset'))
*remove load all together because it's just a wrapper
*keep track ids as they are
pros: simplifies the current api
cons: not very consistent

Option 4 - create a Dataset class
---------------------------------
here's a big discussion about this #225

pro: helps standardize everything
cons: adds complexity, top level api change

nkundiushuti:
besides solving this issues, the dataset object would help with other solutions: sampling, generator. maybe it's worth the effort. we could split it between us and modify the existing loaders.

magdalenafuentes:
Yeah, after discussing a lot yesterday we decided to go for the big change and add the `dataset object`. It will simplify code a lot, increase consistency and change the user API only a bit. We decided to base the implementation on Vincent's great idea in #219.

rabitt:
@nkundiushuti take a look at #296 and let me know what you think about the new proposed API. The Dataset class can of course be extended in the future to include sampling etc as you mention! I started with just porting the existing functionality, and if it seems solid we can add to it

rabitt:
Done in #296

Issue #228, opened 2020-04-08
dataset.choice


rabitt:
An idea from #219 : would it be useful to add a `.choice()` method to each dataset module which returns one random instantiated Track object?

lostanlen:
Yes. This is a nod to `random.choice` https://docs.python.org/3/library/random.html#random.choice

and i suspect it's going to be very useful for machine learning practictioners

It's also good for doing informal demos of mirdata

rabitt:
Implemented in #296

Issue #227, opened 2020-04-08
Search by citation metadata


rabitt:
This came up as an idea from @lostanlen - to be able to search for datasets by metadata from the citation/s, e.g. by author, year etc.

Do we want this feature?

Issue #225, opened 2020-04-08
Should dataset modules have a common Dataset class?


rabitt:
Our current API has the following functions required by every dataset module, and all instances are nearly identical:
* `.download()`
* `.validate()`
* `.track_ids()`
* `.load()`
* `.cite()`

In #219 @lostanlen experimented with adding a `Dataset` class which implements these methods. In addition, he added the following:
* `.readme` (module documentation)
* `.__getitem__`
* `dataset_default_path`
* `.index`
* `.metadata`
* `.choice` for random iteration  (I <3 this idea)

The original decision to not implement a module-level dataset level object was because of usability. In the current API, to, for example, download a dataset:
```python
from mirdata import orchset
orchset.download()
```
Whereas, if download lived inside a dataset object, the API would be:
```python
from mirdata import orchset
orchset_dataset = orchset.Dataset()
orchset_dataset.download()
```

The current solution has the following pros and cons.
Pros:
* the API is intuitive
* The code is relatively easy to follow

Cons:
* We have nearly repeated code in every module
* the inclusion of the 5 methods is only enforced by `test_loaders` + code review.
* There is more code to write (well, copy paste...) when contributing.

Opening this issue to restart the discussion of if a `Dataset` object makes sense, and if so, what the API would look like. I like the idea of generalizing and simplifying the code, as long as it doesn't come at the cost of usability & ease of contribution.

lostanlen:
Thank you for opening this. For the record, #219 was just a wild attempt at unifying module APIs. I didn't see it as a take-it-or-leave-it situation, but more as a conversation starter.

Regarding the example code you wentioned about download, my dream is that the user wouldn't have to import any submodules manually.
So the end goal is not:

```python
from mirdata import orchset
orchset_dataset = orchset.Dataset()
orchset_dataset.download()
```

but really

```python
import mirdata
orchset = mirdata.download("ORCHSET")
```
(we can talk about case sensitivity if you want to allow for `Orchset` and `orchset` aliases)

Where `mirdata.download` would be implemented in some `core.py` Python file.
Here's the (pseudo-)source code to achieve it:

```python
def find_submodule(name, case_sensitive=True):
    if not case_sensitive:
        name = name.lower()
    for submodule_str in mirdata.__all__:
        module_str = "mirdata." + submodule_str
        submodule_name = importlib.import_module(module_str).name
        if not case_sensitive:
            submodule_name = submodule_name.lower()
        if name == submodule.name:
[...]

PR #296, opened 2020-10-20
Dataset object

### LINKED ISSUE #256: move ".cite" string to docstring & return docstring
the current implementation of cite is (e.g. from dali.py)
```python
def cite():
    """Print the reference"""

    cite_data = """
    ===========  MLA ===========
    Meseguer-Brocal, Gabriel, et al.
    "DALI: a large Dataset of synchronized Audio, LyrIcs and notes, automatically created using teacher-student machine
    learning paradigm."
    In Proceedings of the 19th International Society for Music Information Retrieval Conference (ISMIR). 2018.

    ========== Bibtex ==========
    @inproceedings{Meseguer-Brocal_2018,
    Title = {DALI: a large Dataset of synchronized Audio, LyrIcs and notes, automatically created using teacher-student
     machine learning paradigm.},
    Author = {Meseguer-Brocal, Gabriel and Cohen-Hadria, Alice and Peeters, Geoffroy},
    Booktitle = {19th International Society for Music Information Retrieval Conference},
    Editor = {ISMIR}, Month = {September},
    Year = {2018}}
    """
    print(cite_data)
```

Our docs for this method are not very helpful, and it would be great if the reference was itself part of the docs.
image

We could instead move the citation directly to the docstring:
```python
def cite():
    """
    ===========  MLA ===========
    Meseguer-Brocal, Gabriel, et al.
    "DALI: a large Dataset of synchronized Audio, LyrIcs and notes, automatically created using teacher-student machine
    learning paradigm."
    In Proceedings of the 19th International Society for Music Information Retrieval Conference (ISMIR). 2018.

    ========== Bibtex ==========
    @inproceedings{Meseguer-Brocal_2018,
    Title = {DALI: a large Dataset of synchronized Audio, LyrIcs and notes, automatically created using teacher-student
     machine learning paradigm.},
    Author = {Meseguer-Brocal, Gabriel and Cohen-Hadria, Alice and Peeters, Geoffroy},
    Booktitle = {19th International Society for Music Information Retrieval Conference},
    Editor = {ISMIR}, Month = {September},
    Year = {2018}}
    """
    print(cite.__doc__)
```

Alternatively, we could define CITE_DATA as a global in bibtex format, which would get us closer to #227

rabitt:
Implements #293 - this is a breaking, non-backwards compatible change, so we're leaping from 0.2.0 to 0.3.0.

* Abstracts `validate`, `load` (--> `load_tracks`), `track_ids`, `cite` and `download` into a common Dataset class^
* Implements #228 
* would close #256 (it becomes irrelevant)
* Enables #227 in the future
* moves `Track`, `Dataset` and `MultiTrack` base objects into a new module `core.py`
* creats a `mirdata.datasets` submodule for all loaders
* removes `dataset.DATASET_DIR`, sets default folder name to the module name

^This is heavily based on the RFC from #219, and related to #225 !

- [x] Add `Dataset` class
- [x] update usage examples in `README`
- [x] update `CONTRIBUTING`
- [x] update examples in docs
- [x] update all loaders to conform to new API
- [x] update tests
- [x] update test_full_dataset.py
- [x] add tests


rabitt:
@magdalenafuentes - can you take a quick look and let me know if you have any concerns? Note that the downloader has a default, but is configurable if needed, as in `maestro`. It will reduce the lines of code by a LOT but is a breaking change to the top-level API.

Once it looks good to you, i'll start porting the rest over & write some tests!

CONTRIBUTING.md
```diff
@@ -266,97 +280,34 @@ def load_audio(audio_path):
...
-    if data_home is None:
-        data_home = utils.get_default_dataset_path(DATASET_DIR)
-
-    data = {}
-    for key in DATA.index.keys():
-        data[key] = Track(key, data_home=data_home)
-    return data
+    # see download_utils.downloader for basic usage - if you only need to call downloader
+    # once, you do not need this function at all.
+    # ONLY WRITE A CUSTOM FUNCTION IF YOU NEED IT!
```
magdalenafuentes, line 310: hahaha don't shout at people Rachel :P

Also so many deleted lines <3

README.md
```diff
@@ -24,14 +24,15 @@ pip install mirdata
...
+import mirdata
 import random
 
-mirdata.orchset.download()  # download the dataset
-mirdata.orchset.validate()  # validate that all the expected files are there
-orchset_data = mirdata.orchset.load()  # (lazy) load the data index
+orchset = mirdata.Dataset('orchset')
+orchset.download()  # download the dataset
+orchset.validate()  # validate that all the expected files are there
+orchset_data = orchset.load_tracks()  # (lazy) load the data index
```
magdalenafuentes, line 33: remove (lazy)? will see if we remove this completely later

README.md
```diff
@@ -90,31 +91,44 @@ We welcome contributions to this library, especially new datasets. Please see [C
...
 
-### Load the Orchset Dataset
+### Load data for a specific track
 ```python
-import mirdata.orchset
+import mirdata
 
-orchset_data = mirdata.orchset.load()
+orchset = mirdata.Dataset('orchset')
+track = orchset.track('Beethoven-S3-I-ex1')
```
magdalenafuentes, line 113: track -> Track?

magdalenafuentes: or maybe you changed it? if it's a class let's keep it Track (so we have mirdata.Dataset and dataset.Track?)

README.md
magdalenafuentes: also we're changing load --> load_tracks? what happens in the case of recommendation datasets or multitrack? better to keep load?

mirdata/dataset.py
```diff
@@ -0,0 +1,157 @@
+"""Module containing the Dataset base class
+"""
+import importlib
```
magdalenafuentes, line 3: need to add in requirements or not?

README.md
magdalenafuentes: I see, you had to create a `track` in the dataset class... I guess it's good!

mirdata/dataset.py
```diff
@@ -0,0 +1,157 @@
...
+                but missing locally
+            invalid_checksums (list): List of file paths that file exists in the dataset
+                index but has a different checksum compare to the reference checksum
+
+        """
+        missing_files, invalid_checksums = utils.validator(
+            self.index, self.data_home, verbose=verbose
+        )
+        return missing_files, invalid_checksums
+
```
magdalenafuentes, line 157: Beautiful class :'D

mirdata/dataset.py
```diff
@@ -0,0 +1,157 @@
...
+        
+        Returns:
+            track (dataset.Track): an instance of this dataset's Track object
+        """
+        if self.track_object is None:
+            raise NotImplementedError
+        else:
+            return self.track_object(track_id, self.data_home)
+
+    def load_tracks(self):
```
magdalenafuentes, line 77: see my comment above about load()

mirdata/dataset.py
```diff
@@ -0,0 +1,157 @@
...
+        """Load all tracks in the dataset
+
+        Returns:
+            (dict): {`track_id`: track data}
+        
+        Raises:
+            NotImplementedError: If the dataset does not support Track objects
+        """
+        return {self.track(track_id) for track_id in self.track_ids}
```
magdalenafuentes, line 86: maybe add the `tqdm` as in [here](https://github.com/mir-dataset-loaders/mirdata/pull/219/files#diff-8d8cb5987f2bdf3f193ef08ea93956bc4d8d6adf147ba19fd588957114fc4fdfR119-R122)

mirdata/utils.py
```diff
@@ -97,7 +95,7 @@ def check_index(dataset_index, data_home):
     invalid_checksums = {}
 
     # loop over track ids
-    for track_id, track in dataset_index.items():
+    for track_id, track in tqdm.tqdm(dataset_index.items(), disable=not verbose):
```
magdalenafuentes, line 98: <3

magdalenafuentes (commented):
Exciting PR! <3

README.md
rabitt: I was thinking in order to keep everything consistent, we could have `load_tracks`, `load_multitracks` and `load_tables`, which will always be available as methods but might raise an error when dataset doenst have that kind of data.

You can imagine a dataset which has both a table and tracks, or both tracks and multitracks, etc. By breaking it up, everything stays clean

README.md
rabitt: it can be called `Track` instead of `track` but according to my linter it's bad to have a class attribute that is itself a class. 🤷‍♀️ 

I guess it stays a bit cleaner this way too, since `dataset.track` is a function and not a class

mirdata/dataset.py
rabitt: it's in the standard library!

mirdata/dataset.py
rabitt: thanks to Vincent!!

mirdata/dataset.py
rabitt: i thought about it, but it's not slow to load the tracks, since the objects don't load any data.

README.md
magdalenafuentes: Like it!

mirdata/dataset.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # this is a hack to be able to have dataset-specific docstrings
+        self.track = lambda track_id: self._track(track_id)
+        self.track.__doc__ = self._track_object.__doc__  # set the docstring
+
+        # inherit any public load functions from the module
+        for method_name in dir(module):
+            if method_name.startswith("load_"):
+                method = getattr(module, method_name)
+                setattr(self, method_name, method)
```
rabitt, line 70: @magdalenafuentes I don't love this solution... but here's the problem - 
module-level functions weren't part of the dataset object, so if you wanted to use them, you'd have to both create an e.g. `b = mirdata.Dataset('beatles')` object and import the module, like `from mirdata import beatles` to do `beatles.load(b.choice_track.beat_path)`. Which is awful... so I put in this hack, which lists all variables and functions in the module (`dir(module)`) and adds the one starting with `load_` to the class. It works, but it feels very hacky.

The right way i think would be for all the loaders to inherit from a Dataset base class like this one. The top level (new) api could be similar `orchset = mirdata.dataset('orchset')`, but alllll of the loaders would change. The more I write this, the more I think it's the way we should go....and I'll probably work on this tomorrow :)

magdalenafuentes: I see what you mean. I think it makes sense that loaders inherit from this class and not "the other way around". This will be better in the long term as well, people could reuse this class to release their loaders directly with `mirdata-compliable` code

rabitt: I tried it, and it turns out to be much more complicated than it seemed :(. Buckle up for a long story 🙃 

**Option 1: (the one in the current PR) loaders do not extend the Dataset object**
The things i don't like about the current solution in this PR:
1) it's very hacky to import a module inside a class
2) the Track object needs data_home which should be set by the dataset object, so we have to define a function which is a simple wrapper on `Track` in order to call `dataset.track(track_id)` without repeating `data_home`.
3) because of 2, a hack (creating a lambda instead of a method & manually setting its docstring) is necessary for the dataset.track docstring to be specific to the dataset (and this is important to have)
4) manually attaching the module's load methods to the object is hacky

In summary...too many hacks 😸 . What is very nice about the solution in this PR is that adding a new loader is the same  as before (but much less!), and it's relatively clear what's going on. The top-level API is also as we want it.

To remove the hacks, I experimented with having loaders extend the Dataset class. And now the real story beings 📖 

**Option 2: loaders extend Dataset, and the load functions and the Track class are defined inside the Dataset object**
This addresses 1 and 4 but doesn't fix 2 and 3, and in fact makes 2 and 3 even uglier.

See the [dataset-object2 branch](https://github.com/mir-dataset-loaders/mirdata/blob/dataset-object2/mirdata/beatles.py#L215-L266) if you want to see the ugliness of the code. In short, the `Track` object needs access to the `load_*` functions, and `data_home`. Since `Track` is part of the Dataset class, it needs to access it's parent's `self`, which is hackily done in the base Class by [passing self as a variable](https://github.com/mir-dataset-loaders/mirdata/blob/dataset-object2/mirdata/dataset.py#L69-L71) to the `__init__` of `Track`. I looked and it's among the least ugly ways to do this - there are other ways but it's even worse. So now, to initialize a Track you have to call Track(track_id, data_home, dataset_object_self). Super weird, so the user should never see it, so to make it even uglier, we need to hide it by calling it `_Track`. Ugly right? Then, every time contributors are writing a loader, whenever they want to access a `load_*` function inside the `_Track` object, they have to call `self._dataset.load_*` instead of just calling `load_*`. Not the end of the world but not very friendly.

**Option 3: loaders extend Dataset but the load functions and the Track class are defined outside of Dataset**

This is kind of a mid-point between the current PR and Option 1. It would address 1 and 4 in the list of things I don't like, but doesn't solve 2 or 3. It adds two new problems - the users have to manually add the load_* functions to the Dataset object, and instead of defining globals (e.g. `REMOTES`) they have to set hidden variables in the init (e.g `self._remotes`). So it's a bit more copy-paste code than the current solution. 

```python
...
class Dataset(dataset.Dataset):
    def __init__(self, data_home):
       # everything here would be roughly the same as the current PR, but defined as attributes instead of globals
        self.bibtex = "..."
        self._remotes = {
            ...
        }
        ...
       # everything below this would be "extra" stuff the user writes compared to now
        self._track_object = Track
        self.load_beats = load_beats   # manually defined by the user
        self.load_sections = load_sections
        self.load_notes = load_notes
    ...

class Track(track.Track):
    ...

def load_beats(fpath):
    ...

def load_sections(fpath):
    ...

def load_notes(fpath):
    ...
```

Finally, for either of these solutions, we will still have to at some point hackily import a module in order to do something like `mirdata.get_dataset('orchset')` in the API. So we move the hacky import from inside the class to inside a function.

In summary, going back to my list of 4 hacks:
1. is fixed by options 2 or 3, but it moves the hacky import problem elsewhere, and it's not necessarily any better
2 and 3. are not fixed by any option, and are made worse by option 2
4. is fixed by option 2 at the cost of a lot of ugliness. it's also fixed by option 3, but at the cost of manual setting

So on points 1-3, I consider options 1 and 3 equivalent. On point 4, all options are bad in some way.

@magdalenafuentes now that you've read the novel, what do you think? 😂  Option 1, or option 3?

rabitt: and to bias you a bit, I am leaning towards option 1, because despite the hacks, the code is the shortest/cleanest in my opinion

mirdata/dataset.py
```diff
@@ -0,0 +1,172 @@
...
+        if dataset not in DATASETS:
+            raise ValueError(
+                "{} is not a valid dataset in mirdata. Valid datsets are:\n{}".format(
+                    dataset, ",".join(DATASETS)
+                )
+            )
+        module = importlib.import_module("mirdata.{}".format(dataset))
+        self.dataset = dataset
+        self.bibtex = getattr(module, "BIBTEX", "")
+        self._remotes = getattr(module, "REMOTES", {})
```
rabitt, line 49: comment from @drubinstein : make all missing variables None

mirdata/dataset.py
rabitt: discussed with @magdalenafuentes and @drubinstein and we'll go with Option 1 for now, keeping option 3 in mind for the future if we find we need more flexibility. None of these are ideal, so we'll live with the hacks for now.

magdalenafuentes (approved):

docs/source/example.rst
```diff
@@ -22,18 +22,18 @@ Fortunately, we can download Orchset dataset directly.
...
+    orchset.download()
 
 
 Once downloading is done, you can find the the dataset folder.
 
 .. code-block:: bash
     :linenos:
 
-    $ ls ~/mir_datasets/Orchset/
+    $ ls ~/mir_datasets/orchset/
```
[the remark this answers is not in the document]
magdalenafuentes: It's looking great to me!!

magdalenafuentes (approved):

docs/source/example.rst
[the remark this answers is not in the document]
rabitt: I removed the custom `DATASET_DIR` and have the folder name always equal the module name in order to have one less variable. It's a breaking change, but I figured we may as well do it now while we're here.

README.md
nkundiushuti: me too!

mirdata/core.py
```diff
@@ -0,0 +1,232 @@
...
+        """Print the dataset's readme.
+        """
+        print(self._readme_str)
+
+    def cite(self):
+        """Print the reference"""
+        print("========== BibTeX ==========")
+        print(self.bibtex)
+
+    def download(self, partial_download=None, force_overwrite=False, cleanup=True):
```
nkundiushuti, line 145: does this class overwrite the download inside the loader? 
in Saraga we need to adapt the download to query the multi-track audio files individually.

nkundiushuti (approved):

mirdata/core.py
rabitt: for loaders that need a non-standard download method, you can set a `_download` method in the module and it uses that instead. In this PR see for example maestro.py

magdalenafuentes:
Hey @rabitt just FYI we will merge things in this order: dataset object, index version, lfs. I'm working on the index one locally and will create PR once you've merged this one.

rabitt:
Merging this! FYI @genisplaja @PRamoneda and @nkundiushuti - when it comes time to merge the open data loaders I can help port it to the new API, just ask! 

One major change to be aware of - the "DATASET_DIR" where the data actually lives on your computer is now always the same as the name of the module.

lostanlen:
Congratulations @rabitt and all! 🎉

PR #302, opened 2020-10-27
Docs fix 0.2.0

rabitt:
The table of supported datasets on master has a bug:
https://mirdata.readthedocs.io/en/stable/source/datasets.html

This also adds the version to the docs build


rabitt:
merging this now so our docs won't be broken!

PR #303, opened 2020-10-27
Revert "MultiTrack class"

rabitt:
Reverts mir-dataset-loaders/mirdata#295


rabitt:
Reverting this - I realized the new class breaks some conventions that will make #296 more complicated. Updating it now in the original PR!

PR #304, opened 2020-10-28
Multitrack class

rabitt:
Update from the previously merged and reverted #295 

I changed it to have the same basic usage as the `Track` class. This helps us better prepare for 0.3.0

@nkundiushuti sorry for the re-review, hopefully this is easier to integrate into phenix-anechoic!


mirdata/track.py
```diff
@@ -33,11 +33,127 @@ def __repr__(self):
...
+                if sample rates of the tracks are not equal
+                if enforce_length=True and lengths are not equal
+
+        """
+        self._check_mixable()
+        signals = []
+        lengths = []
+        sample_rates = []
+        for k in track_keys:
+            audio, sample_rate = getattr(self.tracks[k], self.track_audio_attribute)()
```
nkundiushuti, line 88: @rabitt I get an error at this line. why is the audio attribute a callable? audio is a property of a Track as far as I understood.
TypeError: 'tuple' object is not callable

this works:
`audio, sample_rate = getattr(self.tracks[k], self.track_audio_attribute)
`

nkundiushuti (commented):
in my case each Track in the MultiTrack is implemented like this, which results in the callable error in the MultiTrack get_targets functions:
`class Track(track.Track):
    """Phenicx-Anechoic Track class

    Args:
        track_id (str): track id of the track
        data_home (str): Local path where the dataset is stored.
            If `None`, looks for the data in the default directory, `~/mir_datasets/Example`


    Attributes:
        instrument (str): instrument corresponding to the track
    """

    def __init__(self, mtrack_id, track_id, data_home=None):

        self.track_id = track_id

        if data_home is None:
            data_home = utils.get_default_dataset_path(DATASET_DIR)

        self._data_home = data_home

        self.audio_path = os.path.join(data_home, DATA.index[mtrack_id][track_id][0])

    @property
    def audio(self):
        """(np.ndarray, float): stereo audio signal, sample rate"""
        return load_audio(self.audio_path)
`

mirdata/track.py
rabitt: ah! yes you're right. Fixing

rabitt:
@nkundiushuti updated - let me know if this works!

nkundiushuti (approved):
it's fine now :)

PR #312, opened 2020-10-30
bugs beatport key dataset loader

### LINKED ISSUE #311: beatport key bad index
I have found an error. Index doesnt have all the keys.

I am sorry.

PRamoneda:
[x] fix the index
[x] improved key getter

Fixes #311


PRamoneda:
I have problems with `to_jams()`

mirdata/beatport_key.py
```diff
@@ -90,8 +90,8 @@ def __init__(self, track_id, data_home=None):
 
     @utils.cached_property
     def key(self):
-        """String: key annotation"""
-        return load_key(self.keys_path)
+        """List of String: list of possible key annotations"""
+        return load_key(self.keys_path).split(' | ')
```
rabitt, line 94: can the `load_key` method do the `.split` there instead of in this property?

PRamoneda: yes of course!

PRamoneda: @rabitt I know that we speak about if to use or not use the keydata structure. I defend that it is overengineering. However now that we are returning a list I think that is better. And I am going to set endtime = 120s and first_time = 0s in every key. I think that is more clear that. the two keys can be possible.

PRamoneda: done here. ba05b89

PRamoneda:
@rabitt I don't like jam key format:
```'pattern': '^N|([A-G][b#]?)(:(major|minor|ionian|dorian|phrygian|lydian|mixolydian|aeolian|locrian))?$'```
I can not adapt it to giantsteps global key format
where more information is added major mode. I think that is easier to use it without keydata.... I comeback to the last commit.... cb59268

rabitt:
> @rabitt I'm not too fond of jam key format:
> 'pattern': '^N|([A-G][b#]?)(:(major|minor|ionian|dorian|phrygian|lydian|mixolydian|aeolian|locrian))?$'
> I can not adapt it to Beatport global key format
> where more information is added to the mode (first two elements are key and mode but later there are other data added for example: G# major mixolydian-b6). Mirdata users must choose what data wants... I think that is easier to use it without keydata.... I comeback to the last commit.... cb59268.
> 
> Giantsteps Key is the same

OK no problem - maybe we can open an issue with Jams to support this in the future, and ignore it in the jam for now, as you did!

mirdata/beatport_key.py
rabitt: What about returning KeyData, as you do in the commit, but just not passing it in the `jams_converter` call?

Also, my point about doing the `.split` within the `load_key` method was so that someone using off-disk data can do
`load_key(custom_keys_path)` and get the same result as if they did `Track.key`

nkundiushuti (approved):

rabitt (changes requested):
one small comment below!

mirdata/beatport_key.py
PRamoneda: I moved the .split to load here: 5688a17  but I reverted it !
I have added it other time! 008739c

PRamoneda: >> What about returning KeyData, as you do in the commit, but just not passing it in the jams_converter call?

I think that there is a huge relationship between giant steps (difficult keys) and Beatport keys (easier keys). The two datasets are to detect keys in EDM mixes... And they have to have consistency between them because people who use one should use the other one. If we can not introduce it in the key space... i think that is better if we keep the format as now (the same format that giant steps)

PRamoneda: What do you think @rabitt ?

rabitt (approved):
👍

Issue #307, opened 2020-10-29
GiantSteps key loader second version


PRamoneda:
Today a new version for giantsteps_key has been uploaded. 
https://zenodo.org/record/4153506 
We have to upload the loader.

nkundiushuti:
what are the changes to match the new data?

rabitt:
Maybe we should wait for #265 to be implemented and then we can use `giantsteps_key` as a first example?

rabitt:
Question for @PRamoneda - what's the difference between the old and the new versions of the datasets? Is it corrections, new tracks, new annotations, etc?

nkundiushuti:
> Maybe we should wait for #265 to be implemented and then we can use `giantsteps_key` as a first example?

yes! this would be a great use case!

PRamoneda:
>>Question for @PRamoneda - what's the difference between the old and the new versions of the datasets? Is it corrections, new tracks, new annotations, etc.?
Hey @rabitt In the first version only was included the key name (tonic + mode Maj/min + diatonic mode). Now it is included a more extended pitch-class as the key tonal centre of the piece. And It is better represented when a key is ambiguous between two keys. I don't if I have explained this better.

magdalenafuentes:
Will be closed by #315 

PR what's the state of this PR? I see that's approved?

PRamoneda:
It is not finished! I have to speak with Angel Faraldo. Last week we would have had a meeting but I was ill :S

PR #313, opened 2020-11-02
second version GiantSteps key

PRamoneda:
https://github.com/mir-dataset-loaders/mirdata/issues/307


PRamoneda:
I have perverted the brach because i havent created it from master.

PR #319, opened 2020-11-04
fix #306 ziplib issue : unzipping archives created on Windows 

### LINKED ISSUE #306: unzip cp437 filename encoding issue
we have a problem in IRMAS because it seems that the archive has been created within Windows with a cp437 encoding. this means that with ziplib some filenames containing utf8 chars (umlauts for instance) aren't decoded properly and these files cannot be read properly. the test_full_dataset fails for irmas at the moment

there are 2 solutions
1. we upload a new version of IRMAS with new zip files
2. we modify unzip and replace extract all to support cp437 encoding for all zip archives

```

def extractall_unicode(zf,zdir):
    for m in zf.infolist():
        data = zf.read(m) # extract zipped data into memory

        if m.filename.encode('cp437').decode() != m.filename.encode('utf8').decode():
            disk_file_name = os.path.join(zdir,m.filename.encode('cp437').decode() )
        else:
            disk_file_name = os.path.join(zdir,m.filename)
    
        dir_name = os.path.dirname(disk_file_name)
        if not os.path.exists(dir_name):
            os.makedirs(dir_name)

        if os.path.isfile(disk_file_name):
            with open(disk_file_name, 'wb') as fd:
                fd.write(data)

def unzip(zip_path, cleanup=True):
    """Unzip a zip file inside it's current directory.

    Args:
        zip_path (str): Path to zip file
        cleanup (bool): If True, remove zipfile after unzipping. Default=False

    """
    zfile = zipfile.ZipFile(zip_path, 'r')
    #zfile.extractall(os.path.dirname(zip_path))
    extractall_unicode(zfile,os.path.dirname(zip_path))
    zfile.close()
    if cleanup:
        os.remove(zip_path)

```

@genisplaja

nkundiushuti:
added tests containing a small archive with unicode filenames


mirdata/download_utils.py
```diff
@@ -181,7 +207,8 @@ def unzip(zip_path, cleanup=True):
 
     """
     zfile = zipfile.ZipFile(zip_path, 'r')
-    zfile.extractall(os.path.dirname(zip_path))
+    # zfile.extractall(os.path.dirname(zip_path))
```
rabitt, line 210: nit: remove

mirdata/download_utils.py
```diff
@@ -172,6 +172,32 @@ def download_zip_file(zip_remote, save_dir, force_overwrite, cleanup=True):
     unzip(zip_download_path, cleanup=cleanup)
 
 
+def extractall_unicode(zfile, out_dir):
+    """Extract all files inside a zip archive to a output directory.
+    In comparison to the zipfile, it checks for correct file name encoding
+
+    Args:
+        zfile (obj): Zip file object created with zipfile.ZipFile
+        out_dir (str): Output fulder
```
rabitt, line 181: nit: `fulder` --> folder

rabitt (approved):
Two nits, but good to go.

In case you didn't already - can you run one or two `test_full_dataset` tests that have zip files to make sure that unzipping is still working on the real data? Some datasets that have zips are `beatport_key`, `orchset`, `rwc_classical`

PR #324, opened 2020-11-05
Update README.md

nkundiushuti:
track_ids is a property. calling it with track_ids results in an error


rabitt (approved):
thanks for catching this!

Issue #329, opened 2020-11-07
Download message is showing {data_home} instead of the actual directory


magdalenafuentes:
```
mirdata.Dataset('medleydb_pitch').download()

    To download this dataset, visit:
    https://zenodo.org/record/2620624#.XKZc7hNKh24
    and request access.

    Once downloaded, unzip the file MedleyDB-Pitch.zip
    and copy the result to:
    {data_home}
```

PR #333, opened 2020-11-09
Fix for issue #329 about datahome printing

genisplaja:
I'm making this PR to try to fix the issue [#329](https://github.com/mir-dataset-loaders/mirdata/issues/329). In my local computer is working well and the pytests are passing correctly, let's see if this proposal fixes the problem!


mirdata/datasets/beatles.py
```diff
@@ -36,17 +36,19 @@
...
         a folder called Beatles with the following structure:
             > Beatles/
                 > annotations/
                 > audio/
-        and copy the Beatles folder to {data_home}
-"""
-
-DATA = utils.LargeData("beatles_index.json")
+        and copy the Beatles folder to {}
+""".format(core.Dataset('beatles').data_home)
```
rabitt, line 51: it works, but it's making the code much slower - just to fill the `data_home` variable, we're instantiating the entire dataset object. is there a way to do this just inside `core`? like keeping the data home variable unset until it's used inside the Dataset object?

rabitt:
I don't know if this will work for sure, but what about something like changing this line - 

https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/core.py#L167

and doing `_download_info.format(data_home)`?

mirdata/datasets/beatles.py
rabitt: I just realized also that this will always set `data_home` to the default, but if someone uses a custom one, it will still print the default

genisplaja:
What about editing this line in `download_utils.py` lines 82 and 83.

```
if info_message is not None:
    print(info_message)
```
And add the the directory name here:
```
if info_message is not None:
    print(info_message.format(save_dir))
```
It works, and I can see in `core.py` that `save_dir` input is `self.data_home`.
Would that be a more feasible solution @rabitt?

rabitt:
Yes that works perfect!

by the way - since you're an admin, feel free to make branches directly in mirdata. Then you won't have use forks if you dont want!

genisplaja:
Ok, thanks!! I will update the changes now for all the datasets with `DOWNLOAD_INFO` and I will commit them.

genisplaja:
I think I know what is happening here, DALI's `DONWLOAD_INFO` requires two string inputs.

rabitt (approved):
👍  thanks @genisplaja !

rabitt:
Fixes #329

PR #334, opened 2020-11-12
remote indexes

PRamoneda:
As we speak on Monday. What do you think?

A REMOTES_INDEX dictionary could be added to each loader that needs a remote index.

### DIFF
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -217,7 +217,7 @@ class cached_property(object):
 
 
 class LargeData(object):
-    def __init__(self, index_file, metadata_load_fn=None, download_path=None):
+    def __init__(self, index_file, metadata_load_fn=None, remote_index=None):
         """Object which loads and caches large data the first time it's
         accessed.
 
@@ -234,16 +234,16 @@ class LargeData(object):
         self._metadata = None
         self.index_file = index_file
         self.metadata_load_fn = metadata_load_fn
-        self.download_path = download_path
+        self.remote_index = remote_index
 
     @cached_property
     def index(self):
-        if self.download_path is not None:
+        if self.remote_index is not None:
             working_dir = os.path.dirname(os.path.realpath(__file__))
             path_index_file = os.path.join(working_dir, "datasets/indexes", self.index_file)
             if not os.path.isfile(path_index_file):
                 path_indexes = os.path.join(working_dir, "datasets/indexes")
-                download_utils.downloader(path_indexes, remotes=self.download_path)
+                download_utils.downloader(path_indexes, remotes=self.remote_index)
         return load_json_index(self.index_file)
 
     def metadata(self, data_home):
--- a/mirdata/utils.py
+++ b/mirdata/utils.py
@@ -29,6 +29,7 @@ import hashlib
 import os
 import json
 import tqdm
+from mirdata import download_utils
 
 
 def md5(file_path):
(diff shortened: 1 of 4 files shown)


mirdata/utils.py
```diff
@@ -233,9 +234,16 @@ def __init__(self, index_file, metadata_load_fn=None):
         self._metadata = None
         self.index_file = index_file
         self.metadata_load_fn = metadata_load_fn
+        self.download_path = download_path
 
     @cached_property
     def index(self):
+        if self.download_path is not None:
```
nkundiushuti, line 241: this should be tested a bit differently: if the index does not exist locally and if self.download_path is not None

nkundiushuti:
so we finally store big indexes in separate zenodo repositories, right?

nkundiushuti:
what is the impact on the tests for the datasets with remote indexes? 
which tests are affected by this change? 
to check: we will probably would not want to download 2GB when we run a test

mirdata/utils.py
```diff
@@ -216,7 +217,7 @@ def __get__(self, obj, cls):
 
 
 class LargeData(object):
-    def __init__(self, index_file, metadata_load_fn=None):
+    def __init__(self, index_file, metadata_load_fn=None, download_path=None):
```
magdalenafuentes, line 220: I guess download_path will be used only in the case of loaders with remote indexes?

mirdata/utils.py
magdalenafuentes: +1

mirdata/utils.py
magdalenafuentes: I'm a bit confused about what is that you're passing to the downloader in `remotes` below. It should be a dictionary with a `RemoteFileMetadata` object but you're passing a path instead? Or I am missing something?

mirdata/utils.py
PRamoneda: Yes!! Tomorrow I wil do the tests!

PRamoneda:
> so we finally store big indexes in separate zenodo repositories, right?

Yes!!

PRamoneda:
> what is the impact on the tests for the datasets with remote indexes?
> which tests are affected by this change?
> to check: we will probably would not want to download 2GB when we run a test

mmm ok I have to mangae it then :S

mirdata/utils.py
PRamoneda: >>I guess download_path will be used only in the case of loaders with remote indexes?
yes, of course!! By default it is not used....

>>I'm a bit confused about what is that you're passing to the downloader in remotes below. It should be a dictionary with a RemoteFileMetadata object but you're passing a path instead? Or I am missing something?
It is a dictionary with a RemoteFileMetadata object or None

magdalenafuentes: gotcha, maybe we can call it `remote_index` or something like that then? using the "_path" I'd expect a `str`

PRamoneda:
> what is the impact on the tests for the datasets with remote indexes?

no impact

> which tests are affected by this change?

none

> to check: we will probably would not want to download 2GB when we run a test

It will be only downloaded if the files have not been downloaded before.

mirdata/utils.py
PRamoneda: done!!!

PRamoneda:
HI!!! It is finished. Codecov errors are not mine.

nkundiushuti (approved):

PR #336, opened 2020-11-17
[WIP] NamedTuples to Classes

### LINKED ISSUE #291: NamedTuples --> Classes
Annotations are currently defined as NamedTuples (e.g. F0Data) or as Track attributes. In order to better support searching & standardization, we've discussed moving them over to Classes, where we can support:
* mir_eval compatibility (#195 )
* data type and unit consistency
* format conversion
* (eventually) automatic JAMS support (related to #316 , #224)

This should be backward compatible.

magdalenafuentes:
Closes #291. 

Converts all `NamedTuples` data types to classes to enforce consistency among loaders and check `mir_eval` compliance.

- [x] Remove all NamedTuples from `utils` and create `classes.py`
- [x] Create functions that check consistency with `mir_eval`
- [x] Convert all `NamedTuples` to `Classes`
- [x] Update tests
- [x] Update `to_jams`
- [x] Fix format inconsistencies
- [x] Fix `groove_midi`
- [x] Complete docs
- [x] Add tests for annotations


magdalenafuentes:
@rabitt let me know what do you think of this structure before I write the other classes

mirdata/classes.py
```diff
@@ -0,0 +1,104 @@
...
+        check_lengths(attributes)
+
+        self.beat_times = beat_times
+        self.beat_positions = beat_positions
+
+
+@dataclass
+class SectionData:
+    """
+    # SectionData = namedtuple("SectionData", ["intervals", "labels"])
```
nkundiushuti, line 58: I have some doubts for the SectionData (maybe similarly for NoteData) 
could you give me an example of annotation? 
how do you represent an interval with an array? do you allow for overlapping intervals then? 
in my case I represented Notes as EventData, because they have a start-time, an end-time and a label.

nkundiushuti:
I left solely a comment but I really like the implementation! great stuff!

magdalenafuentes:
@rabitt I coded most of it, can you take a look and let me know what do you think?

One thing: `groove_midi` had `{'Roland': 'Ride (Bow)', 'General MIDI': 'Ride Cymbal 1', 'Simplified': 'Ride Cymbal (51)'}` as event for each Note, and it should be a list of strings, so I got the `simplified` label only. Thoughts?

magdalenafuentes:
Also @rabitt, I used the `dataclass` wrapper but it's included starting on `Python 3.7`, so the tests in `3.6` are failing. I found a backport [here](https://pypi.org/project/dataclasses/), should we include it to keep supporting `3.6`?

genisplaja:
> @rabitt I coded most of it, can you take a look and let me know what do you think?
> 
> One thing: `groove_midi` had `{'Roland': 'Ride (Bow)', 'General MIDI': 'Ride Cymbal 1', 'Simplified': 'Ride Cymbal (51)'}` as event for each Note, and it should be a list of strings, so I got the `simplified` label only. Thoughts?

@magdalenafuentes About the `groove_midi` issue: The `(51)` is needed in the event annotations? Because if not, I would use `Roland` label as it is more descriptive about the specific drum event. 
For the snare:
```
    38: {
        "Roland": "Snare (Head)",
        "General MIDI": "Acoustic Snare",
        "Simplified": "Snare (38)",
    },
    40: {
        "Roland": "Snare (Rim)",
        "General MIDI": "Electric Snare",
        "Simplified": "Snare (38)",
    },
    37: {
        "Roland": "Snare X-Stick",
        "General MIDI": "Side Stick",
        "Simplified": "Snare (38)",
    }
```
`simplified` here is the same for the whole set of snare events, but you would not distinguish between X-Stick and Rim shots in the event annotation. However, it all depends on the required information in the event annotation!

magdalenafuentes:
@genisplaja good clarification! I'll use `Roland` instead

genisplaja (approved):
That's really great!!! I experimented a bit with it and it looks really nice. Just a couple of questions and comments:

In this PR, for `EventData`, `LyricData` and `KeyData` we are grouping `"start_times"` and `"end_times"` as intervals, when currently we aren't. I can't recall if you commented something about that, the goal is to standarize it with the rest of annotations that use `"intervals"` instead of taking start and end timestamps separatedly?

I tried to implement `DownbeatData` for Saraga loaders (to implement what we discussed at #354), and it looks like it will work quite smoothly and elegantly!

rabitt (commented):
Some comments! We should also add a `tests/test_annotations.py` file

mirdata/classes.py
```diff
@@ -0,0 +1,320 @@
...
+    # customized __init__ to check data types
+    def __init__(self, times, positions=None):
+        attributes = [times, positions]  # list of attributes
+        attributes_names = ["times", "positions"]
+        expected_type = [np.ndarray, np.ndarray]
+        expected_dtype = [float, int]
+
+        # check mir_eval complience
+        check_types(attributes_names, attributes, expected_type, expected_dtype)
+        check_lengths(attributes)
```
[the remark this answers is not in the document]
magdalenafuentes: Mmmm.. The reason I packed that into functions is because I check that in all classes and it will be a lot of repeated code. Also as we want to add more checks later on I think it would be easy to include functions and call them in the different classes when it makes sense

rabitt:
@magdalenafuentes just to not forget - we should merge this with master and include the docs fixes with the new annotation type. Make sure to add mirdata.annotations to source/mirdata.rst so it gets docs.

rabitt:
Some bad news - `dataclasses` is new in python 3.7 and we're supporting python 3.6. I'm switching the classes to regular classes to fix since they're not so different1

mirdata/annotations.py
```diff
@@ -0,0 +1,320 @@
...
+    for name, attribute, exp_type, int_type in zip(
+        attributes_names, attributes, expected_type, expected_dtype
+    ):
+        if attribute is not None:
+            if not isinstance(attribute, exp_type):
+                raise TypeError(
+                    f"{name} should be a {exp_type}, but is a {type(attribute)} instead"
+                )
+            if not np.asarray(attribute).size == 0:  # check list/ndarray not empty
+                if all(isinstance(type(x), int_type) for x in attribute):
```
rabitt, line 23: eek! just found this bug it should be `isinstance(s, int_type)`. When its fixed, there are a lotttt of broken loaders. fixing them now 😱

docs/source/contributing.rst
```diff
@@ -405,17 +405,15 @@ Module example
...
-                end_times.append(float(line[1]))
+                intervals.append([float(line[0]), float(line[1])])
                 annotation.append(line[2])
 
-        annotation_data = utils.EventData(
-            np.array(start_times), np.array(end_times),
-            np.array(annotation))
+        annotation_data = annotations.EventData(
+            np.array(intervals), np.array(annotation)
+        )
```
magdalenafuentes, line 416: Good catch!

PR #342, opened 2020-11-23
Fix validation of datasets without metadata

### LINKED ISSUE #341: Datasets with no general metadata file are not passing test_full_dataset.py
I came across this issue while testing the full Saraga Multitrack Dataset #340.
When testing the full dataset there is an error in this line in the _test_validation_ function: 
```
assert missing_files == {key:{} for key in dataset._index.keys() if not key == "version"}
```

And the error looks like this:
```
AssertionError: assert {'tracks': {}} == {'metadata': {}, 'tracks': {}}
E         Omitting 1 identical items, use -vv to show
E         Right contains 1 more item:
E         {'metadata': {}}
E         Use -v to get the full diff
```

I went to master and tested full dataset for different datasets just to check and I found that this error happens for all the datasets that do not have a general metadata file. I tested:

- Mridangam Stroke (no metadata, failed)
- IRMAS (no metadata, failed)
- beatport Key (no metadata, failed)
- Groove MIDI (has metadata, tests passed).

I also checked that my master branch is up-to-date and it is! Anyone has found this problem?

magdalenafuentes:
Fixes #341


genisplaja:
Thanks for this PR!! I am actually checking it because this morning I tried to change line 144 from utils.py to `if 'metadata' in dataset_index and dataset_index['metadata'] is not None:` just as this PR, because I thought that it is a good idea to keep the same dict structure for all the indexes even if there is no general metadata file, but it failed. Now I'm rerunning it because I think that it should work!

magdalenafuentes:
Great! Thanks :) let me know and if it does I'll merge

genisplaja:
I don't know if you have tried it but in my case isn't working yet (I am not removing `metadata` key from index). I don't know if the problem is when dumping and loading from the JSON index file (converting from None to null and viceversa), but I have debugged the code and tried some alternatives, and when running the code in this PR without deleting the `metadata` key, the error is still there.

Maybe the best idea is to directly remove the `metadata` key from the datasets that do not have general metadata and we specify, in this same PR, in the CONTRIBUTING.md file that the datasets with no general metadata must not include the key in the index.

magdalenafuentes:
Mmmmm... that's strange. It is working for me locally both with `metadata=null` and without. 

Is it not working for the `test_full_dataset` either on `orchset` and `mridangam_stroke`?

magdalenafuentes:
both `orchset` and `mridangam_stroke` pass the full test locally for me

genisplaja:
> Mmmmm... that's strange. It is working for me locally both with `metadata=null` and without.
> 
> Is it not working for the `test_full_dataset` either on `orchset` and `mridangam_stroke`?

I haven't tried in `orchset`, but I tried it for `mridangam_stroke` and it broke. Yeah that's weird! Because I set an `asset` to check that content of `metadata` key was `None` and it was!

genisplaja:
Btw if I remove the `metadata` field from the indexes it works perfectly!

rabitt:
Quick comment - From a code structure perspective, it's better if any null top-level fields are removed, with the exception of version.

genisplaja:
Ok! Understood. So then I guess we can leave the patch as @magdalenafuentes proposed and that would solve the issue. It's working for me now!

genisplaja (approved):

nkundiushuti (approved):

Issue #326, opened 2020-11-06
Maestro breaks if downloading twice and unexpected behaviour download function


magdalenafuentes:
I'm getting this error if I download maestro twice. I don't get this with other datasets (I tried with the `rwc_collection` and `beatport_key`). Even though they don't break, the partial download does different things: in the case of `rwc` it doesn't download a second time, in the case of `beatport` it downloads again.

```
>>> dataset.download(partial_download=['metadata'])
Starting to download ['metadata'] to folder /Users/mf3734/
> downloading metadata
440kB [00:00, 473kB/s]                                                                             
>>> dataset.download(partial_download=['metadata'])
Starting to download ['metadata'] to folder /Users/mf3734/
> downloading metadata
440kB [00:01, 247kB/s]                                                                             
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mf3734/Documents/repos/mirdata/mirdata/core.py", line 169, in download
    cleanup=cleanup,
  File "/Users/mf3734/Documents/repos/mirdata/mirdata/datasets/maestro.py", line 273, in _download
    shutil.move(fpath, save_dir)
  File "/Users/mf3734/anaconda3/envs/mirdata/lib/python3.7/shutil.py", line 564, in move
    raise Error("Destination path '%s' already exists" % real_dst)
shutil.Error: Destination path '/Users/mf3734/maestro-v2.0.0.json' already exists
```

PRamoneda:
I have a similar bug.... I think...

When it is downloading a remote zip if I cancel it (^C) if I come back to download the remote a sha related error shows up :S

PRamoneda:
I think that the best way for solving  my bug is to add 
here https://github.com/mir-dataset-loaders/mirdata/blob/e538acf287514561c4e065c9da918a0c8f65ee28/mirdata/download_utils.py#L134
this code:
```Python
except KeyboardInterrupt:
    os.remove(download_path)
    sys.exit()
```

magdalenafuentes:
@PRamoneda did you check if this was solved by #344? If so we should close this issue

rabitt:
Looking into this:

@PRamoneda 
> When it is downloading a remote zip if I cancel it (^C) and I come back to download the remote a sha related error shows up :S

[...]

PR #344, opened 2020-11-24
fix cancelled download bug

PRamoneda:
As I said on #326 
When it is downloading a remote zip if I cancel it (^C) (or internet breaks down) and I come back to download the remote a sha related error shows up :S. This is because the corrupt .zip (half-downloaded) is not removed


mirdata/download_utils.py
```diff
@@ -22,12 +22,12 @@
...
-    partial_download=None,
-    info_message=None,
-    force_overwrite=False,
-    cleanup=True,
+        save_dir,
+        remotes=None,
+        partial_download=None,
+        info_message=None,
+        force_overwrite=False,
+        cleanup=True,
```
nkundiushuti, line 30: check for black formatting

mirdata/download_utils.py
```diff
@@ -132,7 +132,7 @@ def download_from_remote(remote, save_dir, force_overwrite=False):
                     reporthook=t.update_to,
                     data=None,
                 )
-            except Exception as e:
+            except (urllib.error.URLError, urllib.error.HTTPError) as e:
```
nkundiushuti, line 135: could you add a comment explaining why you do this?
because errors may occur for instance when the hard drive is full it may be better to use OSError (more generic)

nkundiushuti (changes requested):
so when you do download with force_overwrite=True the download works. this solution is for the default case force_overwrite=False when the download is failing for some reason. 

could we also display a message when the checksum for the archives do not match so that the users try to download the dataset with force_overwrite=True?

nkundiushuti (approved):

PRamoneda:
ok done!

magdalenafuentes (approved):

PR #354, opened 2020-12-01
Adding loaders for Saraga Carnatic and Hindustani datasets

genisplaja:
### Description

- [x] Create tests for your loader in `tests/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/datasets.rst`
- [x] Add the module to `mirdata/__init__.py`
- [x] Add the module to the list in the `README.md` file, section `Currently supported datasets`
- [x] Docs building
- [x] Run and pass `test_full_dataset` for Saraga Carnatic
- [x] Run and pass `test_full_dataset` for Saraga Hindustani

#### Please-do-not-edit flag
To reduce friction, we will make commits on top of contributor's pull requests by default unless they use the `please-do-not-edit` flag. If you don't want this to happen don't forget to add the flag when you start your pull request.


nkundiushuti:
@genisplaja  is this ready to review?

genisplaja:
> @genisplaja is this ready to review?

Yes @nkundiushuti, thanks! Sorry I forgot to ask for your reviews! I will do it now.

mirdata/datasets/saraga_carnatic.py
```diff
@@ -152,57 +149,29 @@ def __init__(self, track_id, data_home):
...
-                    'layas': None,
-                }
-
-        # Common attributes for Hindustani and Carnatic tracks
+        # Track attributes
+        metadata = DATA.metadata(self.metadata_path)
+        if metadata is not None and metadata['title'].replace(' ', '_') in self.track_id:
+            self._track_metadata = metadata
+        else:
+            # annotations with missing metadata
```
nkundiushuti, line 157: do we have many files with missing metadata? maybe we can correct them or talk to the people who created the dataset to see what happened.

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+
+    times = []
+    freqs = []
+    with open(pitch_path, 'r') as reader:
+        for line in reader.readlines():
+            times.append(float(line.split('\t')[0]))
+            freqs.append(float(line.split('\t')[1]))
+
+    if not times:
+        return None
```
nkundiushuti, line 303: I think this means that you have to write a test where times=False so this line is triggered

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+            tempo_data.append(sama_interval)
+            matras_per_cycle = line.split(',')[3].split(' ')[1]
+            tempo_data.append(matras_per_cycle)
+            start_time = line.split(',')[4].split(' ')[1]
+            tempo_data.append(start_time)
+            duration = line.split(',')[5].split(' ')[1]
+            tempo_data.append(duration)
+
+            if 'NaN' in tempo_data:
+                return None
```
nkundiushuti, line 363: same thing here: you need to add a test where tempo_data contains NaN

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+    if tempo_path is None:
+        return None
+
+    if not os.path.exists(tempo_path):
+        raise IOError("tempo_path {} does not exist".format(tempo_path))
+
+    tempo_annotation = {}
+    head, tail = os.path.split(tempo_path)
+    sections_path = tail.split('.')[0] + '.sections-manual-p.txt'
+    sections_abs_path = head + '/' + sections_path
```
nkundiushuti, line 336: better use os.path.join or os.path.sep for '/'

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+    intervals = []
+    with open(sama_path, 'r') as reader:
+        for line in reader.readlines():
+            timestamps.append(float(line))
+
+    for i in np.arange(1, len(timestamps)):
+        intervals.append([timestamps[i - 1], timestamps[i]])
+        sama_cycles.append('sama cycle ' + str(i))
+
+    if not intervals:
```
nkundiushuti, line 413: same here

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+                    ]
+                )
+                section_labels.append(
+                    str(line.split(',')[3].split('\n')[0])
+                    + '-'
+                    + str(line.split(',')[1])
+                )
+
+    if not intervals:
+        return None
```
nkundiushuti, line 454: same here

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+        for line in reader.readlines():
+            if len(line.split('\t')) == 4:
+                start_times.append(float(line.split('\t')[0]))
+                end_times.append(
+                    float(line.split('\t')[0]) + float(line.split('\t')[2])
+                )
+                events.append(str(line.split('\t')[3].split('\n')[0]))
+            if len(line.split('\t')) == 3:
+                start_times.append(float(line.split('\t')[0]))
+                end_times.append(
```
nkundiushuti, line 489: same here

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,497 @@
...
+    if not os.path.exists(phrases_path):
+        raise IOError("phrases_path {} does not exist".format(phrases_path))
+
+    start_times = []
+    end_times = []
+    events = []
+    with open(phrases_path, 'r') as reader:
+        for line in reader.readlines():
+            if len(line.split('\t')) == 4:
+                start_times.append(float(line.split('\t')[0]))
```
nkundiushuti, line 482: it looks like 
```
start_times.append(float(line.split('\t')[0]))
                end_times.append(
                    float(line.split('\t')[0]) + float(line.split('\t')[2])
                )
```
is common code to both if branches and what it is different is the events

nkundiushuti (approved):
some test coverage issues might require you test some corner cases (if clauses) in the test_saraga files. I left some other small comments. otherwise it's good to go.

mirdata/datasets/saraga_carnatic.py
genisplaja: In fact, there are no files with missing metadata. I wrote that just in case. Should I remove it?

docs/source/datasets.rst
```diff
@@ -22,94 +22,102 @@ click the link provided in the "Module" column.
...
+| salami_            | Salami              | - audio: ❌         | - :ref:`sections`         | 1359   |
+|                    |                     | - annotations: ✅   |                           |        |
++--------------------+---------------------+---------------------+---------------------------+--------+
+| saraga_carnatic_   | Saraga Carnatic     | - audio: ✅         | - :ref:`tempo`            | 249    |
+|                    |                     | - multichannel: ✅  | - :ref:`phrases`          |        |
+|                    |                     | - annotations: ✅   | - :ref:`phrases`          |        |
+|                    |                     |                     | - :ref:`f0`               |        |
+|                    |                     |                     | - :ref:`f0` (vocal)       |        |
+|                    |                     |                     | - :ref:`sections`         |        |
+|                    |                     |                     | - :ref:`sections` (samas) |        |
```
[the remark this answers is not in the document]
genisplaja: Yes, exactly. I stored the sama annotations as `SectionData` but it could also be stored as `EventData`. For who doesn't know exactly what _sama_ stands for: it's the timestamp of the beginning of a new rhythmic cycle (rhythmic cycles are called _talas_) or the end of the previous cycle. I thought that it would be more understandable for users to store _sama_ timestamps as `SectionData`, but I must say that I was doubtful about storing it as `EventData` as well. What are your thoughts about this @nkundiushuti? 
Sections here are the structural sections of the piece, typically there are between 2 and 4 sections approximately.

mirdata/datasets/saraga_hindustani.py
genisplaja: Ok, sure! I also fixed the line above where I was summing strings to create a path as well.

genisplaja: Ok my bad, the line above should be left like that...

docs/source/datasets.rst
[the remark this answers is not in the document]
nkundiushuti: it's fine to have it as SectionData. maybe we should put  :ref:`sections` (piece) :ref:`sections` (rhythmic,samas)

mirdata/datasets/saraga_carnatic.py
nkundiushuti: I think so, yes. then we will catch any errors in missing metadata and try to correct it.

nkundiushuti (approved):

mirdata/datasets/saraga_carnatic.py
genisplaja: Cool! I'll do that before running the `test_full_dataset` in both loaders.

docs/source/datasets.rst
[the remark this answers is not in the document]
genisplaja: Ok! Yes, this is going to be really helpful!

mirdata/datasets/saraga_hindustani.py
genisplaja: I don't understand what you mean here @nkundiushuti. For both Carnatic and Hindustani collections, melodic phrase annotations are the same. 
First column: start time, second column: flag to identify how representative is the phrase, third column: duration, and fourth column: solfège symbols to identify the phrases.

The length condition here is because when testing the full dataset in the first Saraga version, I found that some melodic phrase annotation rows had no fourth column, or in other words, no annotations of the solfège symbols. **That issue is not happening for the whole annotation file but just in some rows**, that is why I set this condition here, to prevent the `load_phrases` function from crashing, as well as noticing the users that the phrase annotation in certain timestamp is not available.

mirdata/datasets/saraga_carnatic.py
genisplaja: Done. All the tracks have valid metadata!!

genisplaja: Well in order to pass the tests I need to leave this missing metadata conditional. In fact, it is present in every loader. Maybe it can be useful if a metadata file is missing for some reason!

mirdata/datasets/indexes/acousticbrainz_genre_dataset_little_test.json
```diff
@@ -0,0 +1,98 @@
+{
```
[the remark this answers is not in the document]
genisplaja: @PRamoneda should I? I guess this is from your AcousticBrainz remote index implementation.

mirdata/datasets/saraga_hindustani.py
nkundiushuti: Sorry @genisplaja , I didn't explain it well. What I meant is to replace this: 
```
with open(phrases_path, 'r') as reader:
        for line in reader.readlines():
            if len(line.split('\t')) == 4:
                start_times.append(float(line.split('\t')[0]))
                end_times.append(
                    float(line.split('\t')[0]) + float(line.split('\t')[2])
                )
                events.append(str(line.split('\t')[3].split('\n')[0]))
            if len(line.split('\t')) == 3:
                start_times.append(float(line.split('\t')[0]))
                end_times.append(
                    float(line.split('\t')[0]) + float(line.split('\t')[2])
                )
                events.append('No information')
```
with this: 
```
with open(phrases_path, 'r') as reader:
        for line in reader.readlines():
           start_times.append(float(line.split('\t')[0]))
                end_times.append(
                    float(line.split('\t')[0]) + float(line.split('\t')[2])
                )
            if len(line.split('\t')) == 4:
                events.append(str(line.split('\t')[3].split('\n')[0]))
            else:
                events.append('No information')
```

genisplaja: Ok, I see! Now I understand. Thanks!

README.md
```diff
@@ -60,7 +60,8 @@ See the Examples section below for more details, or the [documentation](https://
 * [RWC Jazz](https://staff.aist.go.jp/m.goto/RWC-MDB/rwc-mdb-j.html)
 * [RWC Popular](https://staff.aist.go.jp/m.goto/RWC-MDB/rwc-mdb-p.html)
 * [Salami](https://github.com/bmcfee/salami-data-public/tree/hierarchy-corrections)
-* [Saraga](https://mtg.github.io/saraga/)
+* [Saraga Carnatic](https://mtg.github.io/saraga/)
+* [Saraga Hindustani](https://mtg.github.io/saraga/)
```
magdalenafuentes, line 64: Just checking: is the exact same `url` for both datasets I guess?

mirdata/datasets/indexes/acousticbrainz_genre_dataset_little_test.json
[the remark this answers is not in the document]
magdalenafuentes: I think @nkundiushuti comment is because this file appears to be included "for the first time" in this PR, but if it is part of AcousticBrainz implementation it should be added in that PR instead. Maybe remove it and check if tests are passing, and if so it's better if the files is not in this PR

mirdata/datasets/saraga_carnatic.py
```diff
@@ -621,17 +475,13 @@ def load_phrases(phrases_path):
...
-                    float(line.split('\t')[0]) + float(line.split('\t')[2])
-                )
                 events.append(str(line.split('\t')[3].split('\n')[0]))
-            if len(line.split('\t')) == 3:
-                start_times.append(float(line.split('\t')[0]))
-                end_times.append(
-                    float(line.split('\t')[0]) + float(line.split('\t')[2])
-                )
+            else:
                 events.append('No information')
```
magdalenafuentes, line 485: Maybe an empty event should be simply " "? Just wondering if adding a str there is not misleading

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,499 @@
...
+
+    times = []
+    freqs = []
+    with open(pitch_path, 'r') as reader:
+        for line in reader.readlines():
+            times.append(float(line.split('\t')[0]))
+            freqs.append(float(line.split('\t')[1]))
+
+    if not times:
+        return None
```
magdalenafuentes, line 303: It seems this line is not covered by the tests?

magdalenafuentes: Or was it solved already?

mirdata/datasets/saraga_hindustani.py
```diff
@@ -0,0 +1,499 @@
...
+
+        # Return None if sama file is empty
+        if not timestamps:
+            return None
+
+    for i in np.arange(1, len(timestamps)):
+        intervals.append([timestamps[i - 1], timestamps[i]])
+        sama_cycles.append('sama cycle ' + str(i))
+
+    return utils.SectionData(np.array(intervals), sama_cycles)
```
magdalenafuentes, line 417: I understand from [this paper](https://repositori.upf.edu/bitstream/handle/10230/32893/Srinivasamurthy_ISMIR2017_info.pdf?sequence=1&isAllowed=y) that `sama` is the downbeat, so we should use `BeatData` instead?

magdalenafuentes (commented):
Great monster PR! Good job @genisplaja!

I left some small comments. Another general one, I saw you're loading multiple `mp3` files of around `~4MB` each, can we load less files? Our `test/resources` is increasing rapidly...

README.md
genisplaja: Yes! Should I write it in a single line like:
`[Saraga Carnatic & Hindustani](https://mtg.github.io/saraga/)`

mirdata/datasets/saraga_hindustani.py
genisplaja: Yes! Marius noticed me so I added tests for these exceptions! I think I covered them all, I created an extra test track with files with all these exceptions.

mirdata/datasets/saraga_hindustani.py
genisplaja: Yes, we could! I was using `SectionData` because the annotations are "Sama cycles", instead of beat time-stamps and positions. What I don't clearly see is how to transform Sama annotations to `BeatData`. Should we set the `beat_positions` to 1 for every annotated time-stamp? I guess that would work because at the end of the day, the strings with `sama cycle X` wouldn't be needed.

mirdata/datasets/saraga_carnatic.py
genisplaja: Got it! Yeah I think that's better than adding uneeded confusing strings.

genisplaja:
@magdalenafuentes Regarding the test resources size... I know there is a lot of testing data here, specially with audios, but I need them to be there in order to check that everything works. I tried to select short tracks considering the average length of the dataset songs.
What I can do is shorten them with any RAW to reduce the space usage. I guess there won't be any problem with that! Should I?

Btw, thank you very much for your review and comments!!:)

mirdata/datasets/saraga_hindustani.py
magdalenafuentes: Is a good question, because the cycle information is similar to the `bar number` right? Another option is to include a `DownbeatData` type and include `bar_number`, but it feels redundant. Thinking of #291, actually you could extend the beat class to have bar bar number there no?

magdalenafuentes:
If you can reduce them a bit that's great for now, and let's discuss what we do moving forward

mirdata/datasets/saraga_hindustani.py
genisplaja: I see! So with _extend the beat class_ you mean editting `BeatData` class so it has the `bar_number` attribute, or creating `DownbeatData` in my loader extending `BeatData` with the additional bar?

genisplaja:
Ok, I'll do that! Yes, if we keep adding loaders, test resources will keep growing fast so at some point we will have to discuss about that.

mirdata/datasets/saraga_hindustani.py
magdalenafuentes: I mean that after #336 is merged maybe we can add the `bar_number` more elegantly by inheriting the `BeatData` class and adding the attribute to the child, which would be implemented in the loader. But for now what you can do is to  implement `sama` using `BeatData` and we see if we include the bar number later

genisplaja: Ok! Let's do it like that! I just pushed the updated version using `BeatData`.

README.md
magdalenafuentes: No it's ok like this! Just checking, thanks

magdalenafuentes (approved):

magdalenafuentes:
@genisplaja ready to go!

PR #373, opened 2020-12-18
Using csv for data loading

### LINKED ISSUE #372: Use csv module for data loading
Hey @PRamoneda @genisplaja ! (none of this is urgent, just wanted to open so I don't forget!) 

I noticed this as I was going through some of the loaders and I have a general suggestion. When loading a file spaced by commas, tabs, etc, we should use the `csv` library to load it instead of parsing it manually. 

For example, in saraga_carnatic, you have this code:
```python
def load_phrases(phrases_path):
     ...
    with open(phrases_path, 'r') as reader:
        for line in reader.readlines():
            start_times.append(float(line.split('\t')[0]))
            end_times.append(
                float(line.split('\t')[0]) + float(line.split('\t')[2])
            )
            if len(line.split('\t')) == 4:
                events.append(str(line.split('\t')[3].split('\n')[0]))
            else:
                events.append(' ')
```

it could be replaced by:
```python
def load_phrases(phrases_path):
     ...
    with open(phrases_path, 'r') as fhandle:
        reader = csv.reader(fhandle, delimiter='\t')
        for line in reader:
            start_times.append(float(line[0]))
            end_times.append(
                float(line[0]) + float(line[2])
            )
            if len(line) == 4:
                events.append(str(line[3].split('\n')[0]))
            else:
                events.append(' ')
```

genisplaja:
Partially fixes #372.

Datasets fixed:
- [x] saraga_carnatic
- [x] saraga_hindustani
- [x] beatport_key
- [x] cante100
- [x] irmas
- [x] tonality_classicaldb

From @rabitt - @genisplaja I'll do a quick check through the existing loaders and add a list to the issue!


mirdata/datasets/saraga_carnatic.py
```diff
@@ -289,8 +290,13 @@ def load_tonic(tonic_path):
     if not os.path.exists(tonic_path):
         raise IOError("tonic_path {} does not exist".format(tonic_path))
 
-    with open(tonic_path, 'r') as reader:
-        return float(reader.readline().split('\n')[0])
+    tonic = 0
```
rabitt, line 293: why do you need `tonic=0`?

mirdata/datasets/saraga_carnatic.py
```diff
@@ -347,39 +354,39 @@ def load_tempo(tempo_path):
...
+            tempo_data.append(tempo_bpm)
+            sama_interval = line[2]
+            tempo_data.append(sama_interval)
+            beats_per_cycle = line[3]
+            tempo_data.append(beats_per_cycle)
+            subdivisions = line[4]
+            tempo_data.append(subdivisions)
+
+            if 'NaN' in tempo_data:
+                return None
```
rabitt, line 373: Bonus, if you feel like adding one more test to `test_saraga_carnatic.py` to check this case!

mirdata/datasets/saraga_carnatic.py
```diff
@@ -347,39 +354,39 @@ def load_tempo(tempo_path):
...
+            tempo_apm = line[0]
+            tempo_data.append(tempo_apm)
+            tempo_bpm = line[1]
+            tempo_data.append(tempo_bpm)
+            sama_interval = line[2]
+            tempo_data.append(sama_interval)
+            beats_per_cycle = line[3]
+            tempo_data.append(beats_per_cycle)
+            subdivisions = line[4]
+            tempo_data.append(subdivisions)
```
rabitt, line 370: There's only one line right? I think this whole block can be replaced by:

```python
tempo_data = next(reader)
tempo_apm = tempo_data[0]
tempo_bpm = tempo_data[1]
sama_interval = tempo_data[2]
beats_per_cycle = tempo_data[3]
subdivisions = tempo_data[4]
```

mirdata/datasets/saraga_carnatic.py
```diff
@@ -433,16 +441,17 @@ def load_sections(sections_path):
     intervals = []
     section_labels = []
 
-    with open(sections_path, 'r') as reader:
-        for line in reader.readlines():
+    with open(sections_path, 'r') as fhandle:
+        reader = csv.reader(fhandle, delimiter='\t')
+        for line in reader:
             if line != '\n':
```
rabitt, line 447: I think this will always be false, maybe check instead if line[0] is an empty string?

mirdata/datasets/saraga_carnatic.py
```diff
@@ -470,14 +479,15 @@ def load_phrases(phrases_path):
...
             end_times.append(
-                float(line.split('\t')[0]) + float(line.split('\t')[2])
+                float(line[0]) + float(line[2])
             )
-            if len(line.split('\t')) == 4:
-                events.append(str(line.split('\t')[3].split('\n')[0]))
+            if len(line) == 4:
+                events.append(str(line[3].split('\n')[0]))
             else:
                 events.append(' ')
```
rabitt, line 492: maybe an empty string `""` is better than a string with a space `" "`?

mirdata/datasets/saraga_hindustani.py
```diff
@@ -344,27 +351,28 @@ def load_tempo(tempo_path):
...
             if line != "\n":
-                sections.append(line.split(",")[3].split("\n")[0])
+                sections.append(line[3])
 
     section_count = 0
-    with open(tempo_path, "r") as reader:
-        for line in reader.readlines():
-
+    with open(tempo_path, 'r') as fhandle:
+        reader = csv.reader(fhandle, delimiter=',')
```
rabitt, line 362: same comment as for the other one!

mirdata/datasets/saraga_carnatic.py
genisplaja: True. I was initializing the variable but I guess it's not needed here. Sorry!

mirdata/datasets/saraga_carnatic.py
genisplaja: Cool! I forgot that one. I'll do it together with the rest of the fixes. Thanks for the catch!

mirdata/datasets/saraga_carnatic.py
genisplaja: Yes. This was useful with the other implementation since some annotation files have an additional \n at the end.

mirdata/datasets/saraga_hindustani.py
rabitt: don't forget to update this one the same as you did in the other loader!

rabitt (approved):
Looks amazing! Left one small comment to not forget one change. Otherwise merge at your convenience :)

genisplaja:
> Looks amazing! Left one small comment to not forget one change. Otherwise merge at your convenience :)

Cool! Thank you very much for the feedback:) I'll fix this last comment and also `beatport` and `classicaldb`, and then push the changes and merge in case it's OK.

PR #376, opened 2020-12-18
string to structure

drubinstein:
A work in progress example of:

- Using file-like interfaces instead of passing paths around to the underlying loaders
- Adding type information along the way.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,20 +35,7 @@ show_authors = False
 
 
 # -- Mock dependencies -------------------------------------------------------
-
-# # Mock the dependencies
-from unittest.mock import MagicMock
-
-
-class Mock(MagicMock):
-    @classmethod
-    def __getattr__(cls, name):
-        return MagicMock()
-
-
-MOCK_MODULES = ["librosa", "numpy", "jams", "pretty_midi", "DALI"]
-
-sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+autodoc_mock_imports = ["librosa", "numpy", "jams", "pretty_midi", "DALI"]
 
 
 # # -- General configuration ---------------------------------------------------
--- a/mirdata/datasets/beatles.py
+++ b/mirdata/datasets/beatles.py
@@ -12,6 +12,8 @@
 
 import csv
 import os
+from typing import Optional, Tuple
+
 import librosa
 import numpy as np
 
@@ -19,6 +21,7 @@ from mirdata import download_utils
 from mirdata import jams_utils
 from mirdata import core
 from mirdata import annotations
+from mirdata import io
 
 
 DATA = core.LargeData("beatles_index.json")
@@ -87,38 +90,32 @@ class Track(core.Track):
 
         self._data_home = data_home
         self._track_paths = DATA.index["tracks"][track_id]
-        self.beats_path = core.none_path_join(
-            [self._data_home, self._track_paths["beat"][0]]
-        )
+        self.beats_path = core.none_path_join([self._data_home, self._track_paths["beat"][0]])
         self.chords_path = os.path.join(self._data_home, self._track_paths["chords"][0])
-        self.keys_path = core.none_path_join(
-            [self._data_home, self._track_paths["keys"][0]]
-        )
-        self.sections_path = os.path.join(
-            self._data_home, self._track_paths["sections"][0]
-        )
+        self.keys_path = core.none_path_join([self._data_home, self._track_paths["keys"][0]])
+        self.sections_path = os.path.join(self._data_home, self._track_paths["sections"][0])
         self.audio_path = os.path.join(self._data_home, self._track_paths["audio"][0])
 
         self.title = os.path.basename(self._track_paths["sections"][0]).split(".")[0]
 
     @core.cached_property
-    def beats(self):
+    def beats(self) -> Optional[annotations.BeatData]:
         return load_beats(self.beats_path)
 
     @core.cached_property
-    def chords(self):
+    def chords(self) -> Optional[annotations.ChordData]:
         return load_chords(self.chords_path)
 
     @core.cached_property
-    def key(self):
+    def key(self) -> Optional[annotations.KeyData]:
         return load_key(self.keys_path)
 
     @core.cached_property
-    def sections(self):
+    def sections(self) -> Optional[annotations.SectionData]:
         return load_sections(self.sections_path)
 
     @property
-    def audio(self):
+    def audio(self) -> Tuple[np.ndarray, float]:
         """The track's audio
 
         Returns:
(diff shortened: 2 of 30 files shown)


drubinstein:
@rabitt I updated the PR with what we discussed re: accepting either a file-like object or string representing a path. Does it look good and should I scale it to the rest of mirdata?

mirdata/io.py
```diff
@@ -0,0 +1,48 @@
...
+        elif isinstance(file_path_or_obj, io.StringIO):
+            return func(file_path_or_obj)
+        else:
+            raise ValueError(
+                "Invalid argument passed to {}, argument has the type {}",
+                func.__name__,
+                type(file_path_or_obj),
+            )
+
+    wrapper.__doc__ = func.__doc__
```
psobot, line 25: FWIW [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) could handle this for you.

mirdata/io.py
```diff
@@ -0,0 +1,48 @@
+import io
+from typing import BinaryIO, Callable, Optional, TextIO, TypeVar, Union
+
+T = TypeVar('T')  # Can be anything
+
+
+def coerce_to_string_io(
```
psobot, line 7: Super-pedantic nit: this decorator is called `coerce_to_string_io` but coerces to a file-like object, which `typing` calls `TextIO`, but might not actually pass a `StringIO` to the function it's wrapping.

mirdata/io.py
```diff
@@ -0,0 +1,48 @@
...
+        if not file_path_or_obj:
+            return None
+        if isinstance(file_path_or_obj, str):
+            with open(file_path_or_obj) as f:
+                return func(f)
+        elif isinstance(file_path_or_obj, io.StringIO):
+            return func(file_path_or_obj)
+        else:
+            raise ValueError(
+                "Invalid argument passed to {}, argument has the type {}",
```
psobot, line 20: Would it be useful to include the type(s) expected by the decorator?

mirdata/io.py
```diff
@@ -0,0 +1,48 @@
...
+    def wrapper(file_path_or_obj: Optional[Union[str, TextIO]]) -> Optional[T]:
+        if not file_path_or_obj:
+            return None
+        if isinstance(file_path_or_obj, str):
+            with open(file_path_or_obj) as f:
+                return func(f)
+        elif isinstance(file_path_or_obj, io.StringIO):
+            return func(file_path_or_obj)
+        else:
+            raise ValueError(
```
psobot, line 19: This might be a good case to use [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError):
> Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.

mirdata/io.py
```diff
@@ -0,0 +1,48 @@
...
+                "Invalid argument passed to {}, argument has the type {}",
+                func.__name__,
+                type(file_path_or_obj),
+            )
+
+    wrapper.__doc__ = func.__doc__
+    return wrapper
+
+
+def coerce_to_bytes_io(
```
psobot, line 29: A lot of this code is shared with the above function - would it make sense to share some of this code? (maybe by making the file's mode and the `IO` sub-type parameters?)

mirdata/io.py
psobot: Also documentation on this decorator might be useful - the type annotations do tell the story, but are a bit deeply nested.

psobot (commented):
LGTM with some comments, although I'm not sure I have authority to say "merge it!"

mirdata/io.py
drubinstein: I can try to rewrite it again. Last time I tried it, the results looked kind of messy with a lot of branching.

mirdata/io.py
drubinstein: I'll add a docstring or two

mirdata/io.py
drubinstein: :thu

mirdata/io.py
drubinstein: 👍

mirdata/io.py
drubinstein: Gave it a go.

mirdata/datasets/beatles.py
```diff
@@ -83,38 +86,32 @@ def __init__(self, track_id, data_home):
...
-        )
+        self.keys_path = core.none_path_join([self._data_home, self._track_paths["keys"][0]])
+        self.sections_path = os.path.join(self._data_home, self._track_paths["sections"][0])
         self.audio_path = os.path.join(self._data_home, self._track_paths["audio"][0])
 
         self.title = os.path.basename(self._track_paths["sections"][0]).split(".")[0]
 
     @core.cached_property
-    def beats(self):
+    def beats(self) -> Optional[annotations.BeatData]:
```
rabitt, line 98: quick note about napoleon - we might be able to remove the types from the Cached Properties in the docstring above but we'd have to check

mirdata/datasets/beatles.py
```diff
@@ -141,151 +138,117 @@ def to_jams(self):
...
 
 
-def load_beats(beats_path):
+@io.coerce_to_io(io.Mode.TEXT)
+def load_beats(fbeats: TextIO) -> annotations.BeatData:
     """Load Beatles format beat data from a file
 
     Args:
-        beats_path (str): path to beat annotation file
+        fbeats (str or file-like): path or file-like object pointing to a beat annotation file
```
rabitt, line 161: same comment here - I think the (type annotations) in the docstrings can be removed if we use typing.

mirdata/datasets/beatles.py
```diff
@@ -141,151 +138,117 @@ def to_jams(self):
...
 
     Args:
-        chords_path (str): path to chord annotation file
+        fchords (str or file-like): path or file-like object pointing to a chord annotation file
 
     Returns:
         ChordData: loaded chord data
 
     """
-    if chords_path is None:
```
rabitt, line 204: love that these functions get so much smaller :)

mirdata/datasets/beatles.py
```diff
@@ -141,151 +138,117 @@ def to_jams(self):
...
-    with open(keys_path, "r") as fhandle:
-        reader = csv.reader(fhandle, delimiter="\t")
-        for line in reader:
-            if line[2] == "Key":
-                start_times.append(float(line[0]))
-                end_times.append(float(line[1]))
-                keys.append(line[3])
-
-    key_data = annotations.KeyData(np.array([start_times, end_times]).T, keys)
+    reader = csv.reader(fkeys, delimiter="\t")
```
rabitt, line 219: what happens if `fkeys` is a string and not a file object? I don't think csv.reader can take strings directly

rabitt: oh is this what your decorator function is doing? I'm up for it, but I think it'll be yet another weird mirdata thing that it hard to pick up for new contributors - could this instead call a function that takes a string or file object and returns an object? what do you think is cleaner?

drubinstein: We could change it to that, but then we have to deal with the whole "how do we handle None" business and that makes these functions a little messy again. The reason why I didn't remove the types from the doc string is that it better signifies what the user can pass to the function.

If these functions didn't have to return optional then switching to calling a function (which is what the decorator is kind of doing) would be the same number of lines

drubinstein: My worry would be more for new users than new contributors with the decorator. New contributors should probably find this a bit easier no? With the decorator would make it so they never have to worry about file handling. They only have to worry about handling the contents of a file.

rabitt: True! I opened #398 to make this decorator and others clear for contributors.

drubinstein:
@rabitt I think this is ready for final review.

docs/conf.py
```diff
@@ -35,20 +35,7 @@
...
-class Mock(MagicMock):
-    @classmethod
-    def __getattr__(cls, name):
-        return MagicMock()
-
-
-MOCK_MODULES = ["librosa", "numpy", "jams", "pretty_midi", "DALI"]
-
-sys.modules.update((mod_name, Mock()) for mod_name in MOCK_MODULES)
+autodoc_mock_imports = ["librosa", "numpy", "jams", "pretty_midi", "DALI"]
```
rabitt, line 38: ❤️

mirdata/datasets/salami.py
```diff
@@ -190,36 +192,28 @@ def __init__(self, track_id, data_home):
...
     @core.cached_property
-    def sections_annotator_2_uppercase(self):
-        if self.sections_annotator2_uppercase_path is None:
-            return None
+    def sections_annotator_2_uppercase(self) -> Optional[annotations.SectionData]:
         return load_sections(self.sections_annotator2_uppercase_path)
 
     @core.cached_property
-    def sections_annotator_2_lowercase(self):
-        if self.sections_annotator2_lowercase_path is None:
```
rabitt, line 212: 👍

mirdata/datasets/tonality_classicaldb.py
```diff
@@ -181,109 +186,81 @@ def to_jams(self):
...
+def load_spectrum(fhandle: TextIO) -> np.ndarray:
     """Load Tonality classicalDB spectrum data from a file
 
     Args:
-        spectrum_path (str): path to spectrum file
+        fhandle: File-like object or path to spectrum file
 
     Returns:
-        np.array: spectrum data
+        np.narray: spectrum data
```
rabitt, line 229: np.ndarray?

mirdata/datasets/tonality_classicaldb.py
```diff
@@ -181,109 +186,81 @@ def to_jams(self):
...
 
 
-def load_musicbrainz(musicbrainz_path):
+@io.coerce_to_string_io
+def load_musicbrainz(fhandle: TextIO) -> Dict[Any, Any]:
     """Load Tonality classicalDB musicbraiz metadata from a file
 
     Args:
-        musicbrainz_path (str): path to musicbrainz metadata  file
+        fhandle: File-like object or path to musicbrainz metadatafile
```
rabitt, line 257: metadata file

tests/test_loaders.py
```diff
@@ -186,13 +174,9 @@ def test_download(mocker, httpserver):
                 url = dataset.remotes[key].url
                 try:
                     request = requests.head(url)
-                    assert request.ok, "Link {} for {} does not return OK".format(
-                        url, dataset_name
-                    )
+                    assert request.ok, "Link {} for {} does not return OK".format(url, dataset_name)
```
rabitt, line 177: we've got different line lengths set on black 😬

tests/test_loaders.py
```diff
@@ -46,9 +46,7 @@
 
 
 def create_remote_index(httpserver, dataset_name):
-    httpserver.serve_content(
-        open(REMOTE_DATASETS[dataset_name]["local_index"], "rb").read()
-    )
+    httpserver.serve_content(open(REMOTE_DATASETS[dataset_name]["local_index"], "rb").read())
```
rabitt, line 49: just to be sure - did you change any code in this file? it looks like this is all just reformatting

rabitt (commented):
Some docs nits and a few questions. I love how much this simplifies a lot of the code!

mirdata/datasets/beatles.py
```diff
@@ -145,149 +148,115 @@ def to_jams(self):
...
     """Load a Beatles audio file.
 
     Args:
-        audio_path (str): path to audio file
+        fhandle: path or file-like object pointing to an audio file
 
     Returns:
-        * np.ndarray - the mono audio signal
-        * float - The sample rate of the audio file
+        np.ndarray: the mono audio signal
```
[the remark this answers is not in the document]
drubinstein: That explains it.

mirdata/datasets/cante100.py
```diff
@@ -320,92 +322,82 @@ def to_jams(self):
...
-            freqs.append(float(line[1]))
+    reader = csv.reader(f0, delimiter=",")
+    for line in reader:
+        times.append(float(line[0]))
+        freqs.append(float(line[1]))
 
     times = np.array(times)
     freqs = np.array(freqs)
-    confidence = (freqs > 0).astype(float)
+    confidence = (cast(np.ndarray, freqs) > 0).astype(float)
```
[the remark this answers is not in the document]
drubinstein: It's for typing. Because freqs is a List and then converted to an np.array, mypy still thinks it is its original typing (List). `cast`does nothing but help the type checker understand the type has changed. The other option is to assign np.array(freqs) to a different variable.

mirdata/datasets/giantsteps_tempo.py
```diff
@@ -198,59 +200,46 @@ def to_jams_v2(self):
...
+def load_tempo(fhandle: TextIO) -> annotations.TempoData:
     """Load giantsteps_tempo tempo data from a file ordered by confidence
 
     Args:
-        tempo_path (str): path to tempo annotation file
+        fhandle: File-like object or path to tempo annotation file
 
     Returns:
-        list: list of annotations.TempoData
+        annotations.TempoData: Tempo data
```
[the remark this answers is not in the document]
drubinstein: Yea. If you go to the return statement you'll see `return annotations.TempotData(...)` not `return [annotaitons.TempoData(), ...]`

tests/test_loaders.py
drubinstein: This is all just reformatting.

tests/test_loaders.py
drubinstein: I think this is cause you only run black on the source, not tests. I can add an extra commit to have black run on the tests directory via tox.ini to this PR if you want.

mirdata/datasets/beatles.py
```diff
@@ -145,149 +148,115 @@ def to_jams(self):
...
 
 
-def load_audio(audio_path):
+@io.coerce_to_bytes_io
+def load_audio(fhandle: BinaryIO) -> Tuple[np.ndarray, float]:
     """Load a Beatles audio file.
 
     Args:
-        audio_path (str): path to audio file
+        fhandle: path or file-like object pointing to an audio file
```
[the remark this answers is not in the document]
drubinstein: prepare to have your mind blown
[I can put it in the description apparently with an autodoc setting](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints)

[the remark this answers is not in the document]
drubinstein: Screen Shot 2021-01-13 at 20 56 57

and it looks bad...

mirdata/datasets/beatles.py
```diff
@@ -145,149 +148,115 @@ def to_jams(self):
         )
 
 
-def load_audio(audio_path):
+@io.coerce_to_bytes_io
+def load_audio(fhandle: BinaryIO) -> Tuple[np.ndarray, float]:
```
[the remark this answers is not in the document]
drubinstein: Okay so I found the underlying issue and made a couple of changes here. Looks like readthedocs was installing their default version of sphinx (1.8.5). By upgrading sphinx to 3.4.0 it now renders correctly like on my laptop.

mirdata/datasets/cante100.py
```diff
@@ -320,92 +322,82 @@ def to_jams(self):
...
-        raise IOError("spectrogram_path {} does not exist".format(spectrogram_path))
-    parsed_spectrogram = np.genfromtxt(spectrogram_path, delimiter=" ")
+    parsed_spectrogram = np.genfromtxt(faudio, delimiter=" ")
     spectrogram = parsed_spectrogram.astype(np.float)
 
     return spectrogram
 
 
-def load_audio(audio_path):
+def load_audio(faudio: str) -> Tuple[np.ndarray, float]:
```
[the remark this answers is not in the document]
drubinstein: I put a blurb in the contributing section if that's okay.

mirdata/datasets/beatles.py
```diff
@@ -145,149 +148,115 @@ def to_jams(self):
...
     """Load a Beatles audio file.
 
     Args:
-        audio_path (str): path to audio file
+        fhandle: path or file-like object pointing to an audio file
 
     Returns:
-        * np.ndarray - the mono audio signal
-        * float - The sample rate of the audio file
+        * np.ndarray: the mono audio signal
```
rabitt, line 159: This is still not rendering properly.

image

Can you keep them as they were?

mirdata/datasets/beatles.py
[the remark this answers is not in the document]
rabitt: I think it's better if the docs are consistent across functions/loaders. 

```
fhandle (str or TextIO): path or file-like object pointing to an audio file
```

mirdata/datasets/beatles.py
drubinstein: Ah rats I messed up. I'll change it to dash

mirdata/datasets/rwc_classical.py
```diff
@@ -315,8 +312,8 @@ def _position_in_bar(beat_positions, beat_times):
         beat_times (np.ndarray): raw rwc time stamps
 
     Returns:
-        * np.ndarray: normalized beat positions
-        * np.ndarray: normalized time stamps
+        * np.ndarray - normalized beat positions
```
rabitt, line 315: thanks!

rabitt (commented):
Two itty bitty nits and then lets merge!

mirdata/datasets/cante100.py
```diff
@@ -320,92 +322,82 @@ def to_jams(self):
...
     """Load a cante100 audio file.
 
     Args:
-        audio_path (str): path to audio file
+        fhandle (str): path to an audio file
 
     Returns:
-        * np.ndarray - the mono audio signal
-        * float - The sample rate of the audio file
+        * np.ndarray: the mono audio signal
```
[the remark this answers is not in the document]
drubinstein: done

rabitt (approved):

PR #377, opened 2020-12-18
Better docs for datasets

rabitt:
* Updates the Dataset object, and each loader extends the Dataset object.
* Adds a top-level `mirdata.load(dataset)` method
* automatically generates mirdata.DATASETS list
* moves some functions from utils to core (e.g. LargeData, cached property) -- this fixes a previous circular import between `utils` and `download_utils`.
* moves remaining functions in utils to a new module called `validate`


rabitt:
This is working only for beatles.py - take a look at the code/docs and see what you think. The code for everything remains the same, but we have to add a Dataset instance at the bottom of every loader. The good thing is the docs & api are much more clear.

If we go this way, we would also maybe want to add a top level function, e.g. `mirdata.get_dataset('beatles')` to save the user from having to do `from mirdata.datasets import beatles` and then `dataset = beatles.Dataset`.

mirdata/datasets/beatles.py
```diff
@@ -286,3 +285,36 @@ def _fix_newpoint(beat_positions):
...
+            data_home,
+            index=DATA.index,
+            name="beatles",
+            track_object=Track,
+            bibtex=BIBTEX,
+            remotes=REMOTES,
+            download_info=DOWNLOAD_INFO,
+        )
+
+    @core.copy_docs(load_audio)
```
magdalenafuentes, line 306: This is a bit obscure IMO, either we go with something more explicit (if possible)  or will have to add a clear explanation in the docs :/

magdalenafuentes (commented):
Nice! I agree that is nice that each dataset in the docs now has all the functionalities (we should keep in mind how to colapse the info at some point in the docs, if possible). Only minor comment on the docs inheritance which is a bit obscure for contributors right now IMO. Also +100 to the get_dataset funcion

mirdata/datasets/beatles.py
rabitt: I agree it's obscure... I couldn't find a better way to include the load functions along with the docs. The only more explicit way to do it is to copy paste the docs from each loader function :(. I'll try to make a clear explanation in the contributing

magdalenafuentes: Yeah, I couldn't think of an alternative now either. I'm good with keeping it like this and adding comments on `CONTRIBUTING.md`. Should we add comments on @cached_property as well?

tests/test_loaders.py
```diff
@@ -217,151 +191,174 @@ def test_download(mocker, httpserver):
...
-            dataset = mirdata.Dataset(dataset_name, data_home=data_home)
-            dataset_default = mirdata.Dataset(dataset_name, data_home=None)
+            dataset = module.Dataset(data_home)
+            dataset_default = module.Dataset()
+        else:
+            continue
+            # TODO - fix the dataset object to work with remote index
+            # remote_index = create_remote_index(httpserver, dataset_name)
+            # dataset = module.Dataset(data_home, index=remote_index)
+            # dataset_default = module.Dataset(index=remote_index)
```
magdalenafuentes, line 235: This is done already right?

tests/test_loaders.py
```diff
@@ -217,151 +191,174 @@ def test_download(mocker, httpserver):
...
-                    dataset.track("~faketrackid~?!")
-                continue
+            dataset = module.Dataset(data_home)
+            dataset_default = module.Dataset()
+        else:
+            continue
+            # TODO - fix the dataset object to work with remote index
+            # remote_index = create_remote_index(httpserver, dataset_name)
+            # dataset = module.Dataset(data_home, index=remote_index)
+            # dataset_default = module.Dataset(index=remote_index)
```
magdalenafuentes, line 304: Same comment as above

magdalenafuentes (commented):
Looks good to me! I really like the change of splitting `utils` into `core` and `validate`, makes sense. See comments about `mirdata.load(dataset)` and let me know what do you think!

magdalenafuentes:
It seems it's reducing coverage but I only saw one line missing from tests (L448 `core.py`)

README.md
```diff
@@ -27,7 +27,7 @@ pip install mirdata
 import mirdata
 import random
 
-orchset = mirdata.Dataset('orchset')
+orchset = mirdata.load('orchset')
```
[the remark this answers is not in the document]
rabitt: thinking about this. I totally agree with `load` being perceived maybe as loading to memory. trying to come up with a solution. `get` is a bit vague, and sounds like download a bit. `dataset` is confusing, because we already have `Dataset` the object and `datasets` the module. I'm worried about adding a third `dataset` the function.

I think i'll go with `initialize`? Unless anyone has another idea!

tests/test_loaders.py
rabitt: not anymore...I re-broke it 🙃 the tests dont work and they need to be updated. It works for the end-user but it's no longer tested

README.md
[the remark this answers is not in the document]
genisplaja: Yeah, I think `initialize` is a good idea! I'll help to avoid confusion with the word "dataset" around. Unless anyone proposes a different name, I'll arrange it on my docs PR to save time!

rabitt:
I'm going to go ahead and merge this, since it actually will only very minimally conflict with #379 which is waiting for some small changes anyway.

PR #399, opened 2021-01-07
remove line audio size in gs tempo test track

PRamoneda:
This is a recurrent error.

In CircleCi librosa works nice. However, In other systems, sometimes the size is a bit different.

If someone has a cleaner proposal, we are open minded :D


magdalenafuentes:
So you're testing locally and the size of the file is not right? Just sanity check: are you sure is the same exact audio file?

PRamoneda:
Yes! @genisplaja dicussed this topic with rachel in one meeting

magdalenafuentes:
It happens in other loaders as well?

magdalenafuentes (approved):
If you discussed this with @rabitt already let's merge since is a current fail

rabitt:
I get an error for this test locally as well, coming from slight differences in different mp3 decoders. I'm ok with removing it!

rabitt (approved):

PR #403, opened 2021-01-07
Fix RWC empty files and index

### LINKED ISSUE #402: RWC classical breaks with empty annotation files and a few are not included in the index
Running the complete tests as in #390, noticed that the `rwc_classical` loader breaks when trying to load sections from an empty file (there are empty annotation files in the dataset!). I will fix this by checking and returning `None` in those cases.

Also, noticed another issue: some files have a single audio file but multiple annotation files, which are right now ignored in the index. 

Will check other RWC loaders in case, added them in #390.

magdalenafuentes:
Closes #402


magdalenafuentes:
Locally I'm having the issue reported in #399, so this should be merged after that PR is done

rabitt (approved):
Looks great! Quick note - I'm opening an issue about some inconsistencies in how we handle missing annotations. In short - i think missing/empty annotations should be None instead of e.g. SectionData initialized with None, but it's not always what we do!

PR #404, opened 2021-01-07
Fix beatport decoding issue

genisplaja:
Fixes one of the issues found in #390 for beatport_key dataset loader. It was an issue related to encoding/decoding of special characters (accents, etc).
Full dataset tested for macOS and I am about to test it in Ubuntu.


genisplaja:
The `test_full_dataset` in Ubuntu is passing as well!

scripts/legacy/make_beatport_key.py
```diff
@@ -4,7 +4,7 @@
 import os
 
 
-beatport_key_INDEX_PATH = '../mirdata/indexes/beatport_key_index.json'
+beatport_key_INDEX_PATH = '../../mirdata/datasets/indexes/beatport_key_index.json'
```
rabitt, line 7: can you move this script out of scripts/legacy and into just scripts since you updated it to the new api? :)

scripts/legacy/make_beatport_key.py
```diff
@@ -42,7 +46,7 @@ def make_beatport_key_index(data_path):
             else:
                 meta = (meta_path.replace(data_path + '/', ''), md5(meta_path))
 
-            beatport_key_index[track_id] = {
+            beatport_key_index['tracks'][track_id] = {
```
rabitt, line 49: how did the unicode issue in the index file above get fixed? It looks like nothing fundamental changed in this script that would affect the unicode - is the json file above reproducible?

genisplaja: Basically, the error was in the index itself. I think that at some point, the special characters got encoded wrongly for some reason. I checked that the actual encoding matches with the correct Unicode encoding! 
I ran the full test on my local MacBook pro machine and on our Linux server and it passed successfully for both!

scripts/legacy/make_beatport_key.py
genisplaja: Sure! I forgot to change it afterward. Thanks for the note!!

scripts/legacy/make_beatport_key.py
rabitt: I guess my question is - if we reran this script to create the index again, would it have the correct encoding? When you regenerated it this time was it the output of this script?

genisplaja: Yes, the actual index in this PR was generated by the script in this PR. 
However, I cannot assure you that if this script is run on a different machine, the encoding will be the same. I can try to add some lines to enforce this encoding system!

PRamoneda (approved):

Issue #401, opened 2021-01-07
List of available datasets in code?


magdalenafuentes:
To initialize a dataset the user needs to know the name of the dataset loader in `mirdata`, is that easily accessible anywhere? The names are not in the [table of available datasets](https://mirdata.readthedocs.io/en/latest/source/quick_reference.html).

genisplaja:
That's true! We could add it into the table of datasets or even create some sort of `mirdata.dataset_list()` method in the core file to print out the whole list of names to initialize the datasets.

rabitt:
That's true! It's only in the API documentation. Definitely +1 for a `mirdata.list_datasets()` method. I'll make a PR for it!

rabitt:
So it's actually available through `mirdata.DATASETS`, but we all forgot about it 😂  so I'm making a small helper function to make it more clear + obvious in the docs

magdalenafuentes:
Haha great :)

PR #407, opened 2021-01-08
Add a list_datasets function

rabitt:
Implements #401 
* adds a top-level `list_datasets` function
* adds tests for functions in `initialize.py`
* moves docs for "Initialize" above Dataset Loaders, and outside of Core


mirdata/initialize.py
```diff
@@ -5,6 +5,15 @@
...
 
+def list_datasets():
+    """Get a list of all mirdata dataset names
+
+    Returns:
+        list: list of dataset names as strings
+    """
+    return DATASETS
+
+
```
magdalenafuentes, line 16: I find a bit confusing to do `mirdata.initialize.list_datasets` and also `mirdata.initialize(dataset)`. Also for some reason in the docs we're getting a lot of `mirdata.list_datasets.__dir__()`, `mirdata.list_datasets.__new__()` which we probably don't won't there?

I'd suggest we either simply use `mirdata.DATASETS`, or we move `list_datasets()` to `__init__.py` (i understand it's ok to have code there as in any other module?). I'm a bit more inclined for the second. Also, should we use `list_datasets` or `list_loaders`?

In any case, what about adding in the docs a small code-block at the top of `overview` and `tutorial` to show people they can easily see what loaders are available?

magdalenafuentes (approved):
Nice! :) good to go!

PR #411, opened 2021-01-09
fix tonality classicldb full dataset with download

PRamoneda:
I have a problem without  --skip-download. I don't know as I created the dataset at the same time the loader... My fault.
It is fixed. After this PR, every full dataset test are working


rabitt:
@PRamoneda just to make sure I understand what the problem was - the links to the dataset download were wrong?

PRamoneda:
yes, I don't know why. :(

rabitt (approved):

PR #418, opened 2021-01-14
fix typo from "RWC classical" to "RWC popular"

ooyamatakehisa:


docs/source/table.rst
```diff
@@ -184,7 +184,7 @@
      - 50
      - :rwc:`\ `
 
-   * - RWC classical
+   * - RWC popular
```
magdalenafuentes, line 187: Good catch! Thanks :)

magdalenafuentes (approved):

PR #421, opened 2021-01-14
formatting

drubinstein:
This PR handles a few bits related to formatting:
1. python3.8 is now tested for and is used as the basepython for black
2. tox has been refactored to include a `check-formatting` and `format` env
3. `check-formatting` has been added to circleci


rabitt (approved):
🖤

PR #423, opened 2021-01-15
Add mypy (not strict)

drubinstein:
This PR adds a mypy check but not strict meaning that it only checks what has already been added. I fixed up some typing along the way.

### DIFF
--- a/mirdata/datasets/tonality_classicaldb.py
+++ b/mirdata/datasets/tonality_classicaldb.py
@@ -141,19 +141,19 @@ class Track(core.Track):
         self.title = self.audio_path.replace(".wav", "").split("/")[-1]
 
     @core.cached_property
-    def key(self) -> str:
+    def key(self) -> Optional[str]:
         return load_key(self.key_path)
 
     @core.cached_property
-    def spectrum(self) -> np.ndarray:
+    def spectrum(self) -> Optional[np.ndarray]:
         return load_spectrum(self.spectrum_path)
 
     @core.cached_property
-    def hpcp(self) -> np.ndarray:
+    def hpcp(self) -> Optional[np.ndarray]:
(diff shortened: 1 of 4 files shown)


mirdata/datasets/tonality_classicaldb.py
```diff
@@ -141,19 +141,19 @@ def __init__(self, track_id, data_home):
         self.title = self.audio_path.replace(".wav", "").split("/")[-1]
 
     @core.cached_property
-    def key(self) -> str:
+    def key(self) -> Optional[str]:
```
magdalenafuentes, line 144: Hey Dave, thanks for this! I have a naive/dumb question: `no-strict` means that `mypy` doesn't check the use of `None`, or am I confused? If that's what it does, why to use `Optional` here?

drubinstein: It doesnt check the use of None. Mypy does not run at runtime. It analyzes the types passed in the editor to ensure inputs/outputs match. In this case the Use of Optional says that `key` will either return `str` or `None` (`Optional` is really `Union[str, None]`)

[strict](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict) in this case would apply a bunch of checks which are hard to run until types are spread to the rest of the code base.

`load_key(self.key_path)` returns an `Optional[str]` (courtesy of the coerce_to_string_io decorator). The typing was previously correct so this addition fixes the typing. Thanks to mypy I found the problem in the first place.

magdalenafuentes: I see, thanks @drubinstein!

rabitt (approved):
This is great! Thanks @drubinstein

rabitt:
@drubinstein feel free to merge at your convenience

PR #436, opened 2021-01-25
Simplify track object + update contributing docs

### LINKED ISSUE #432: Message of invalid track IDs in Track object shouldn't require manually adding dataset name
Or if it does require the dataset name, it should do it automatically to reduce code in the loaders and have one less thing to add when contributing

rabitt:
Creates an `__init__` in `core.Track` which sets many of the boilerplate attributes we are currently copy-pasting. This changes how we (internally) call `Track` objects within Dataset objects, but this is hidden from the end user.

* Closes #432 
* moves `_load_metadata` functions to `_metadata` within `Dataset` objects. Removes metadata from `LargeData` object'
* updates tests to test using `dataset.track` instead of `module.Track`
* updates contributing to match above
* fixes small formatting issues in docs
* updates how metadata is loaded in saraga datasets (previously was treated as global, but there are track-specific metadata files)
* disables all remote index tests (see issue #437)

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -200,8 +200,8 @@ You may find these examples useful as references:
 * `A simple, fully downloadable dataset <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/tinysol.py>`_
 * `A dataset which is partially downloadable <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/beatles.py>`_
 * `A dataset with restricted access data <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/medleydb_melody.py#L33>`_
-* `A dataset which uses global metadata <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/tinysol.py#L114>`_
-* `A dataset which does not use global metadata <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/gtzan_genre.py#L36>`_
+* `A dataset which uses dataset-level metadata <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/tinysol.py#L114>`_
+* `A dataset which does not use dataset-level metadata <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/gtzan_genre.py#L36>`_
 * `A dataset with a custom download function <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/maestro.py#L257>`_
 * `A dataset with a remote index <https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/acousticbrainz_genre.py>`_
 
@@ -530,9 +530,10 @@ Track Attributes
 Custom track attributes should be global, track-level data.
 For some datasets, there is a separate, dataset-level metadata file
 with track-level metadata, e.g. as a csv. When a single file is needed
-for more than one track, we recommend using writing a ``_load_metadata`` method
-and passing it to a ``LargeData`` object, which is available globally throughout 
-the module to avoid loading the same file multiple times.
+for more than one track, we recommend using writing a ``_metadata`` cached property
+in the Dataset object (see the dataset module example code above). When this is specified,
+it will populate a tracks hidden `_track_metadata` field, which can be accessed from
+the Track object.
 
 Load methods vs Track properties
 --------------------------------
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -530,10 +530,36 @@ Track Attributes
 Custom track attributes should be global, track-level data.
 For some datasets, there is a separate, dataset-level metadata file
 with track-level metadata, e.g. as a csv. When a single file is needed
-for more than one track, we recommend using writing a ``_metadata`` cached property
+for more than one track, we recommend using writing a ``_metadata`` cached property (which
+returns a dictionary, either keyed by track_id or freeform)
 in the Dataset class (see the dataset module example code above). When this is specified,
-it will populate a tracks hidden `_track_metadata` field, which can be accessed from
+it will populate a track's hidden ``_track_metadata`` field, which can be accessed from
 the Track class.
+
+For example, if ``_metadata`` returns a dictionary of the form:
+
+.. code-block:: python
+
+    {
+        'track1': {
+            'artist': 'A',
+            'genre': 'Z'
+        },
+        'track2': {
+            'artist': 'B',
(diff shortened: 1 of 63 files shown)


mirdata/datasets/acousticbrainz_genre.py
```diff
@@ -397,12 +371,23 @@ class Dataset(core.Dataset):
...
+        self,
+        data_home=None,
+        remote_index=None,
+        remote_index_name=None,
+    ):
+        if remote_index and remote_index_name:
+            data = core.LargeData(remote_index_name, remote_index=remote_index)
+            index = data.index
+        else:
+            index = None
```
magdalenafuentes, line 384: Not covered by tests apparently

mirdata/datasets/acousticbrainz_genre.py
```diff
@@ -453,7 +438,7 @@ def download(self, partial_download=None, force_overwrite=False, cleanup=False):
                 )
                 if os.path.isdir(first_dir_path):
                     file_downloaded = True
-                    print(
+                    logging.info(
```
magdalenafuentes, line 441: Not covered :/ I guess this was like this before? Maybe ignore for now (I opened #440 to cover this stuff later)

mirdata/datasets/cante100.py
```diff
@@ -417,13 +339,72 @@ def __init__(self, data_home=None):
...
+        releases = [release.text for release in root.iter("anthology")]
+
+        # duration
+        minutes = [float(minute.text) * 60 for minute in root.iter("duration_m")]
+        seconds = [float(second.text) for second in root.iter("duration_s")]
+        durations = [m + s for (m, s) in zip(minutes, seconds)]
+
+        metadata = dict()
+        metadata["data_home"] = self.data_home
+        for i, j in zip(indexes, range(len(artists))):
```
magdalenafuentes, line 397: This comment probably doesn't belong here, but at this point I'm understanding that for track-level metadata we’re assuming that the metadata at dataset-level is a dictionary? And this is not enforced in the code but left to the contributor right? i.e. the core functionality checks if the `track_id` is in `metadata` so the contributor should write the code accordingly

magdalenafuentes: So is some sort of metadata index?

mirdata/datasets/dali.py
```diff
@@ -126,54 +114,39 @@ class Track(core.Track):
...
+        self.ground_truth = self._track_metadata.get("ground-truth")
+        self.artist = self._track_metadata.get("artist")
+        self.title = self._track_metadata.get("title")
+        self.dataset_version = self._track_metadata.get("dataset_version")
+        self.scores_ncc = self._track_metadata.get("scores", {}).get("NCC")
+        self.scores_manual = self._track_metadata.get("scores", {}).get("manual")
+        self.album = self._track_metadata.get("metadata", {}).get("album")
+        self.release_date = self._track_metadata.get("metadata", {}).get("release_date")
+        self.genres = self._track_metadata.get("metadata", {}).get("genres")
+        self.language = self._track_metadata.get("metadata", {}).get("language")
```
magdalenafuentes, line 149: I like a lot how this simplifies code <3

tests/test_loaders.py
```diff
@@ -465,8 +475,10 @@ def test_multitracks(httpserver):
         if dataset_name not in REMOTE_DATASETS:
             dataset = module.Dataset()
         else:
-            remote_index = create_remote_index(httpserver, dataset_name)
-            dataset = module.Dataset(index=remote_index)
+            # TODO fix these tests
+            continue
+            # remote_index = create_remote_index(httpserver, dataset_name)
+            # dataset = module.Dataset(remote_index=remote_index)
```
magdalenafuentes, line 481: we have quite a few tests to fix 😅

docs/source/contributing.rst
```diff
@@ -530,9 +530,10 @@ Track Attributes
 Custom track attributes should be global, track-level data.
 For some datasets, there is a separate, dataset-level metadata file
 with track-level metadata, e.g. as a csv. When a single file is needed
-for more than one track, we recommend using writing a ``_load_metadata`` method
-and passing it to a ``LargeData`` object, which is available globally throughout 
-the module to avoid loading the same file multiple times.
+for more than one track, we recommend using writing a ``_metadata`` cached property
+in the Dataset class (see the dataset module example code above). When this is specified,
+it will populate a tracks hidden `_track_metadata` field, which can be accessed from
```
magdalenafuentes, line 535: I think we should make explicit that `metadata` has to be an index-like dictionary when track-level information is available, maybe add a small `metadata` section explaining the cases or something like that? I'm afraid that because `_track_metadata` is hidden it might be a bit obscure for the contributor if we don't explain more.

magdalenafuentes (approved):
Great PR! I like a lot how the loaders' code is being simplified!

My only comment is on transparency for the contributor: I'll add a more detailed (still small)  explanation either in the docs or in the example (or both) to make sure they understand when they have to do a `metadata_index`, and what is done with the metadata: propagated to the track-level when needed. One thing I'm not sure I understood is what happens when metadata is not a dictionary, it is not propagated to the track right? It stays in the dataset level? Or am I confused? Sorry for the verbosity! Great work!

mirdata/datasets/acousticbrainz_genre.py
rabitt: many of the lines of acousticbrainz_genre need to be checked because of the remote index. I opened #437 to address it

mirdata/core.py
```diff
@@ -271,6 +278,49 @@ class Track(object):
...
+
+        self._data_home = data_home
+        self._track_paths = index["tracks"][track_id]
+
+        if metadata and track_id in metadata:
+            self._track_metadata = metadata[track_id]
+        elif metadata:
+            self._track_metadata = metadata
+        else:
+            self._track_metadata = {}
```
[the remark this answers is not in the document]
rabitt: good point - changing it! Worth mentioning - remember that before we let missing metadata fail gracefully, because the loaders wouldn't initialize if it wasn't yet downloaded? Now it's not so much of a problem, and I changed it to raise an error if the metadata is accessed but the file isn't there.

mirdata/datasets/cante100.py
rabitt: it's admittedly a bit hacky - 90% of the metadata files are dictionaries mapping track ids to metadata, but some are not (e.g. ikala, which is keyed by "singer id"), which is why we check if the track id is a key or not.

In cante it falls into the "regular" case, where the key is the track id.

tests/test_loaders.py
rabitt: 😬

magdalenafuentes (approved):

PR #444, opened 2021-01-28
fix black command

PRamoneda:
Genis and I have had the same problem in the last days. Black command in contributing is not the correct one.

new command on contributing page:

```bash
black --target-version py36 --target-version py37 --target-version py38 --skip-string-normalization mirdata/ tests/
```

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -392,7 +392,7 @@ If github shows a red ``X`` next to your latest commit, it means one of our chec
 
 ::
 
-    black --target-version py36 --target-version py37 --target-version py38 --skip-string-normalization mirdata/ tests/
+    black --target-version py38 mirdata/ tests/
 
 2. the test coverage is too low -- this means that there are too many new lines of code introduced that are not tested.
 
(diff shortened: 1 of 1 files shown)


docs/source/contributing.rst
```diff
@@ -392,7 +392,7 @@ If github shows a red ``X`` next to your latest commit, it means one of our chec
 
 ::
 
-    black --target-version py37 --skip-string-normalization mirdata/
+    black --target-version py36 --target-version py37 --target-version py38 --skip-string-normalization mirdata/ tests/
```
genisplaja, line 395: For me, the only way I was able to pass the formatting CircleCI tests was by running the black formatting without the `--skip-string-normalization` flag. 
Is this supposed to be an error?

PRamoneda: I don't know :S

rabitt: Good catch, and sorry for the troubles it caused!

I think it should be simply:
```black --target-version py38 mirdata tests```
meaning that we no longer want to skip the string normalization, and we'll have the formatting target just python 3.8.
(it's how it's running now in `tox.ini` as of #421 )

nkundiushuti (approved):

rabitt (approved):

rabitt:
@PRamoneda feel free to merge!

Issue #440, opened 2021-01-26
Increase test coverage


magdalenafuentes:
There are some lines of code here and there that are not covered, we should at some point include them in the tests as well to be closer to 100%.

I'm dropping this here so it becomes a task in the future :)

PR #460, opened 2021-02-02
Increase track attribute test coverage

### LINKED ISSUE #459: acousticbrainz_genre attributes fail
Accessing several of the attributes in acousticbrainz_genre raise `KeyErrors`, including `artist`, `title`, and `album`.

rabitt:
Fixes #459 

In the process of fixing it, i've tightened up our tests (partially addresses #440 ), which revealed a few more small bugs.

In addition, I tweaked the Track class to lazy-load dataset metadata. Instead of accessing Dataset._metadata when the class is initialized, it's accessed when an attribute requiring it is accessed. I added tests to check that things are lazy loading properly as well!

### DIFF
--- a/tests/test_loaders.py
+++ b/tests/test_loaders.py
@@ -291,6 +291,10 @@ def test_track():
             dataset.track("~faketrackid~?!")
 
 
+# This tests the case where there is no data in data_home.
+# It makes sure that the track can be initialized and the
+# attributes accessed, but that anything requiring data
+# files errors (all properties and cached properties).
 def test_track_placeholder_case():
     data_home_dir = "not/a/real/path"
 
--- a/tests/test_loaders.py
+++ b/tests/test_loaders.py
@@ -10,7 +10,7 @@ import requests
 
 import mirdata
 from mirdata import core, download_utils
-from tests.test_utils import DEFAULT_DATA_HOME
+from tests.test_utils import DEFAULT_DATA_HOME, get_attributes_and_properties
 
 DATASETS = mirdata.DATASETS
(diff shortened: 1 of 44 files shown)


tests/test_loaders.py
```diff
@@ -278,6 +291,42 @@ def test_track():
             dataset.track("~faketrackid~?!")
 
 
+def test_track_placeholder_case():
```
magdalenafuentes, line 294: Maybe add explanation, from your own words XD:

This function tests the case where there is no data in data_home. It makes sure that the track can be initialized and the attributes accessed, but that anything requiring data files errors (all properties and cached properties).

tests/test_saraga_carnatic.py
```diff
@@ -58,13 +58,13 @@ def test_track():
...
-        "audio-violin": (np.ndarray, float),
-        "audio-vocal-s": (np.ndarray, float),
-        "audio-vocal": (np.ndarray, float),
+        # "audio-mix": (np.ndarray, float),
+        # "audio-ghatam": (np.ndarray, float),
+        # "audio-mridangam-left": (np.ndarray, float),
+        # "audio-mridangam-right": (np.ndarray, float),
+        # "audio-violin": (np.ndarray, float),
+        # "audio-vocal-s": (np.ndarray, float),
+        # "audio-vocal": (np.ndarray, float),
```
magdalenafuentes, line 67: why are this ones commented?

magdalenafuentes (approved):
Great fix, ready to go with two minor comments

PR #479, opened 2021-03-09
add readthedocs config file

rabitt:
Warnings in the docs build fail silently, and result in parts of the docs not rendering (e.g. the table of supported datasets). This should switch it, so that docs warnings cause an error in the build, and will show up in our test integrations


rabitt:
FYI @nkundiushuti @PRamoneda @genisplaja - once this is merged, we'll easily know when there are problems with the docs, because the build will fail.

magdalenafuentes (approved):
♥️

PR #490, opened 2021-04-07
Marius/fixunzip487

### LINKED ISSUE #487: bug in extractall_unicode
@PRamoneda is working on good sounds dataset and the archive comes in a format which causes an error:
UnicodeEncodeError: ‘charmap’ codec can’t encode character ‘\uf00d’ in position 62: character maps to <undefined>

this is the function where the error appears:
```
def extractall_unicode(zfile, out_dir):
    """Extract all files inside a zip archive to a output directory.
    In comparison to the zipfile, it checks for correct file name encoding
    Args:
        zfile (obj): Zip file object created with zipfile.ZipFile
        out_dir (str): Output folder
    """
    for m in zfile.infolist():
        data = zfile.read(m)  # extract zipped data into memory
        if m.filename.encode("cp437").decode() != m.filename.encode("utf8").decode():
            disk_file_name = os.path.join(out_dir, m.filename.encode("cp437").decode())
        else:
            disk_file_name = os.path.join(out_dir, m.filename)
        dir_name = os.path.dirname(disk_file_name)
        if not os.path.exists(dir_name):
            os.makedirs(dir_name)
        if not os.path.isdir(disk_file_name):
            with open(disk_file_name, "wb") as fd:
                fd.write(data)
```
If we do this:
`m.filename.encode("cp437", errors="ignore").decode()`
then we do not have the error anymore
so there should be some weird char in a filename in the archive.

nkundiushuti:
I refactored extractall_unicode to fix #487


rabitt:
@nkundiushuti let me know when this is ready for review! And thinking it might be worth adding some more edge test cases that trigger the new error that you found in test_download.py

nkundiushuti:
> @nkundiushuti let me know when this is ready for review! And thinking it might be worth adding some more edge test cases that trigger the new error that you found in test_download.py

ready to review! I found an old laptop with windows 7 and created an archive which can trigger more tests.

mirdata/download_utils.py
```diff
@@ -8,6 +8,7 @@
 import tarfile
 import urllib
 import zipfile
+import chardet
```
rabitt, line 11: is this already a dependency of one of the other libraries? We don't explicitly specify it in setup.py

mirdata/download_utils.py
```diff
@@ -246,10 +247,28 @@ def extractall_unicode(zfile, out_dir):
     for m in zfile.infolist():
         data = zfile.read(m)  # extract zipped data into memory
 
-        if m.filename.encode("cp437").decode() != m.filename.encode("utf8").decode():
-            disk_file_name = os.path.join(out_dir, m.filename.encode("cp437").decode())
+        ### get filename
```
rabitt, line 250: style nit - can you use one `#` for comments?

mirdata/download_utils.py
```diff
@@ -246,10 +247,28 @@ def extractall_unicode(zfile, out_dir):
...
+        except UnicodeEncodeError:
+            ### utf encoding
+            filename = name.encode("utf8")
+
+        ### check for irmas filename encoding
+        if filename.decode(errors="ignore") != name.encode("utf8").decode():
+            filename = name.encode("cp437")
+
+        ### detect encoding
+        encoding = chardet.detect(filename)["encoding"]
```
rabitt, line 264: I'm a bit confused by the logic - is this needed? It looks like we should know the encoding based on the result of the try/except

mirdata/download_utils.py
```diff
@@ -246,10 +247,28 @@ def extractall_unicode(zfile, out_dir):
     for m in zfile.infolist():
         data = zfile.read(m)  # extract zipped data into memory
 
-        if m.filename.encode("cp437").decode() != m.filename.encode("utf8").decode():
-            disk_file_name = os.path.join(out_dir, m.filename.encode("cp437").decode())
+        ### get filename
+        name = m.filename
+        try:
```
rabitt, line 252: I have to admit I'm pretty confused by this code! So we're...taking the original filename, encoding it according to different schemes (lines 252-261), then detecting the encoding we just used (line 264), and then decoding it again?

I have no better suggestion... but is there not a more straightforward way?

rabitt (approved):
Some little comments. If you don't see another way, feel free to merge this :)

mirdata/download_utils.py
nkundiushuti: it's a dependency of an existing library. should we put it in setup.py?

rabitt: I think so, yes, just in case one of those libraries someday remove this dependency!

mirdata/download_utils.py
nkundiushuti: the original solution was this:
```
 if m.filename.encode("cp437", errors="ignore").decode(errors="ignore") != m.filename.encode("utf8", errors="ignore").decode(errors="ignore"):
            disk_file_name = os.path.join(out_dir, m.filename.encode("cp437", errors="ignore").decode(errors="ignore"))
        else:
            disk_file_name = os.path.join(out_dir, m.filename)
```
but the tests fail for this because ignoring errors in encoding leads to an incorrect encoding for irmas. 
so here:
```
        try:
            # non-utf encoding
            filename = name.encode("cp437")
        except UnicodeEncodeError:
            # utf encoding
            filename = name.encode("utf8")

        # check for irmas filename encoding
        if filename.decode(errors="ignore") != name.encode("utf8").decode():
            filename = name.encode("cp437")
```
I deal with exceptions related to irmas and good-sounds
while here:
```
        # detect encoding
        encoding = chardet.detect(filename)["encoding"]
        # decode with the encoding and ignore errors in filename
        if encoding is not None:
            filename = filename.decode(encoding, errors="ignore")
        else:
            filename = filename.decode(errors="ignore")
```
I try to deal with the other encodings

nkundiushuti: I will run tests with a few datasets to make sure this does not break anything.

nkundiushuti: unless we know exactly the encoding then this is difficult. I have tried to refactor this code but the tests break with any other solution.

nkundiushuti: @rabitt , I found a more elegant solution to this problem and refactored the code.

mirdata/download_utils.py
```diff
@@ -243,13 +244,22 @@ def extractall_unicode(zfile, out_dir):
...
-            disk_file_name = os.path.join(out_dir, m.filename.encode("cp437").decode())
-        else:
-            disk_file_name = os.path.join(out_dir, m.filename)
+        filename = m.filename
+
+        # if block to deal with irmas and good-sounds archives
+        # check if the zip archive does not have the encoding info set
+        if m.flag_bits & ZIP_FILENAME_UTF8_FLAG == 0:
+            # encode-decode filename only if it's different than the original name
+            if filename.encode("cp437").decode(errors="ignore") != filename:
```
rabitt, line 257: nit: this could be a single if with an `and`

mirdata/download_utils.py
```diff
@@ -243,13 +244,22 @@ def extractall_unicode(zfile, out_dir):
         out_dir (str): Output folder
 
     """
+    ZIP_FILENAME_UTF8_FLAG = 0x800
```
rabitt, line 247: love the refactor! it reads much better now

nkundiushuti: thanks for the review! will merge now

PR #491, opened 2021-04-08
Add non-standard case instructions to contributing docs

### LINKED ISSUE #488: Loaders with optional dependencies should check and print message if missing
For loaders such as `dali` that have optional dependencies, right now if the dependency is not installed it breaks and is obscure to the user. We should instead check if the dependency is there, and if it is not to print a message such `WARNING: this loader needs a dependency that is not installed by mirdata by default. Please run pip install X`

rabitt:
Adds/updates instructions for some common non-standard cases (not freely downloadable data, loaders needing additional python dependencies, and datasets with more than one version).

Additionally, updates example.py in the contributing docs to match our current workflow.

Would close #488


docs/source/contributing_examples/example.py
```diff
@@ -182,6 +193,10 @@ def to_jams(self):
         # -- see the documentation for `jams_utils.jams_converter for all fields
 
 
+# -- this decorator allows this function to take a string or an open bytes file as input
+# -- and in either case converts it to an open file handle.
+# -- It also checks if the file exists
+# -- and, if None is passed, None will be returned 
 @io.coerce_to_bytes_io
 def load_audio(fhandle):
     """Load a Example audio file.
```
genisplaja, line 202: Cool! Thas was definitely needed.

Should we also include here an example of the usage of `typing`? e.g:
```
def load_audio(fhandle: BinaryIO) -> Tuple[np.ndarray, float]:
```
or 
```
def beats(self) -> Optional[annotations.BeatData]:
```
We use it a lot in the loaders but not in the example files.

genisplaja (approved):
Hello @rabitt, this is a great update for the docs! I just left a couple of comments:)

docs/source/contributing_examples/example.py
rabitt: Yes good point! adding it

Issue #489, opened 2021-04-06
Support multiple versions/samples


rabitt:
Right now we don't easily support multiple versions. After an offline discussion we plan to:

* update index names (dataset_version_index.json)
* update index path, supporting backward compatibility (in dataset structure) for v1 datasets in current location
* add a version argument to `Dataset`
* set default versions for existing dataset

rabitt:
Closed by #496

PR #496, opened 2021-04-17
Multiple version support

rabitt:
Addresses #489 

* allows multiple versions via multiple index files
* updates how we handle remote indexes
* updates downloader to download remote indexes, if present
* updates documentation


docs/source/contributing_examples/test_example.py
```diff
@@ -8,7 +8,7 @@
 def test_track():
     default_trackid = "some_id"
     data_home = "tests/resources/mir_datasets/dataset"
-    dataset = example.Dataset(data_home)
+    dataset = example.Dataset(data_home, version="test")
```
magdalenafuentes, line 11: I'm not sure I follow why this is needed here...

mirdata/datasets/indexes/giantsteps_key_index_1.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "+",
+  "version": "1.0",
```
magdalenafuentes, line 2: Does this change match the version in Zenodo?

mirdata/datasets/indexes/gtzan_genre_index_1.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": null,
+  "version": 1.0,
```
magdalenafuentes, line 2: The index version should be the same as the dataset version right? If the dataset doesn't have an official version, should be change it to `1.0` or simply leave the index as `gtzan_genre_index.json`?

mirdata/datasets/indexes/ikala_index_1.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": null,
+  "version": 1.0,
```
magdalenafuentes, line 2: same comment

mirdata/download_utils.py
```diff
@@ -56,9 +57,12 @@ def downloader(
...
             A dictionary of RemoteFileMetadata tuples of data in zip format.
             If None, there is no data to download
+        index (core.Index):
+            A mirdata Index class, which may contain a remote index to be downloaded
+            or a subset of remotes to download by default.
         partial_download (list or None):
             A list of keys to partially download the remote objects of the download dict.
-            If None, all data is downloaded
+            If None, all data specified by the index is downloaded
         info_message (str or None):
```
magdalenafuentes, line 66: With this phrasing I understand that the index contains the `remote index` (if any) AND the `remote data` to download, but that's not the case right?

mirdata/download_utils.py
```diff
@@ -71,14 +75,27 @@ def downloader(
     if not os.path.exists(save_dir):
         os.makedirs(save_dir)
 
+    if not index:
```
magdalenafuentes, line 78: Does this mean that we have to specify the index in the donwload function?

mirdata/download_utils.py
```diff
@@ -71,14 +75,27 @@ def downloader(
...
+
     if cleanup:
         logging.warning(
             "Zip and tar files will be deleted after they are uncompressed. "
             + "If you download this dataset again, it will overwrite existing files, even if force_overwrite=False"
         )
 
-    if remotes is not None:
-        if partial_download is not None:
+    if index.remote:
```
magdalenafuentes, line 87: I'm sorry I think I'm being slow, but I'm confused by which this index is, because the `indexes` under `datasets/indexes` are unchanged right? And they don't have a `remote` key but neither do the `INDEXES` dictionary in the loader? Am I completely lost?

magdalenafuentes (commented):
Great that we're gonna support multiple versions :D! Thanks @rabitt!!

I left a few comments mainly to understand a few things before I continue with the review. 

One silly question to make sure: if we update the indexes names, when upgrading the version of `mirdata` the user would not notice a difference right? All indexes will be updated locally for them as well?

rabitt:
Thanks for the comments @magdalenafuentes !

> One silly question to make sure: if we update the indexes names, when upgrading the version of mirdata the user would not notice a difference right? All indexes will be updated locally for them as well?

Right - the index names will get updated with the library, so should be backward compatible.

docs/source/contributing_examples/test_example.py
rabitt: For the tests, it uses the "test" index, instead of the "default" one. Right now, in all cases except acousticbrainz, the test and default versions are the same. But for acousticbrainz, the test version is the small local index, whereas the default one is more than 1 GB and remote.

mirdata/datasets/indexes/giantsteps_key_index_1.0.json
rabitt: whoops you're right! Fixing it

mirdata/datasets/indexes/gtzan_genre_index_1.0.json
rabitt: I was hesitating about this too. I changed them all to be 1.0 for consistency, but not sure what the right answer is. I was thinking that even if they don't have an official version, it is still the "first" version. What do you think?

mirdata/download_utils.py
rabitt: I created a new `core.Index` class, which is just a container of information for indexes. It has three important attributes:
```
* remote : None (default) or a RemoteFileMetadata object, saying where to download the index from
* partial_download : None (default) or a list of keys of `REMOTES`, in case the index only requires a subset of the data.
* get_path(): returns the path to the index. The function looks to see if remote is None or not to know if the indexes in the usual spot, or somewhere else.
```

So all that to answer your question - the index might contain a remote index (which should get downloaded with the rest of the data), and might also specify a subset of remote data. I was imagining e.g. a dataset with two versions of annotations - if each version is a separate download, you might want to only download the version of the annotations you're using for that version, rather than everything.

mirdata/download_utils.py
rabitt: Yes, now it's a required argument of the downloader. It could be optional, but I made it mandatory for now to avoid future bugs in case we update an indexes to be remote - this way we don't have to remember to also update the downloader.

mirdata/download_utils.py
rabitt: Right - the indexes under datasets/indexes are unchanged.
Maybe the confusing thing is that there's `INDEXES` which is a dictionary, and the values are `core.Index` objects. The input to the downloader is a single `core.Index` object, not `INDEXES`. `The `Index` objects all have `.remote` attributes, but most of them are set to `None`. Right now only acousticbrainz_genre has an index with .remote set.

docs/source/contributing_examples/test_example.py
magdalenafuentes: Gotcha, so you want developers to test the remote indexes in a transparent way. Right now tests for `acousticbrainz` run sneakily using the small index but is not transparent, am I getting this right?

mirdata/datasets/indexes/gtzan_genre_index_1.0.json
magdalenafuentes: Yeah, I had the same question when I introduced the version in the index for the first time... I didn't want to "make up" a version in case that was confusing, but I'm not sure if it matters to much since it's the only version this datasets have anyway

mirdata/download_utils.py
magdalenafuentes: I see, that's great! If nothing is specified in remotes, does `mirdata` download all versions?

magdalenafuentes: dismiss, you already answer that offline in the meeting

mirdata/download_utils.py
magdalenafuentes: from the conversation offline I understood that the `initialize` function is the one to set the index version, and that propagates to the rest of the code. That is smooth from the point of view of the user. Did I get it right?

magdalenafuentes:
I checked the docs you added and they look good! Though I didn't find where it explains the use of remotes in `core.Index`? also, should `core.Index` be under `core`? I didn't see it 

nit: the menu in the docs is looking funny, it looks like an alignment issue probably introduced by another PR accidentally, maybe we can fix that here?

mirdata/download_utils.py
rabitt: yes exactly

mirdata/datasets/indexes/gtzan_genre_index_1.0.json
rabitt: OK, out of laziness/convenience, I'll leave them this way (null -> 1.0)

rabitt:
> nit: the menu in the docs is looking funny, it looks like an alignment issue probably introduced by another PR accidentally, maybe we can fix that here?

I don't see it - can we do it in another PR?

> Though I didn't find where it explains the use of remotes in core.Index? 

It's in the special cases, under "large indexes"

> also, should core.Index be under core? I didn't see it

You're right I forgot to add docs to core.Index! Added now

PR #498, opened 2021-04-22
Adding loader for Slakh

### LINKED ISSUE #263: Add Slakh
I'd like to add [Slakh](http://www.slakh.com/) to mirdata.

@ethman @pseeth - would you guys be able to help review if i do? Also, I've seen your companion repo [slackh-utils](https://github.com/ethman/slakh-utils) - it looks to me like the repo was mostly used to create the dataset. Is there functionality there that would make it worth having it as a dependency of mirdata?

rabitt:

- [x] Create tests for your loader in `tests/datasets/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/quick_reference.rst`

Will close #263


rabitt:
@ethman Can I bother you for a review? :) Let me know if I confused or missed anything!

ethman:
Yup! I'll try to dig in within the next few days. Thanks for doing this!

rabitt:
@ethman Thanks! Outside of the implementation, I had a few doubts:
- There are stems with no midi/audio files. I read in your documentation that you don't render "unusual" sounds. Am I correctly assuming that the midi file for the mixtures also exclude those unusual sources? If that's the case, do you think it's worth including them in the index at all?
- If you happen to have a local copy of slakh2100-redux, could you try:

mirdata/io.py
```diff
@@ -47,3 +53,71 @@ def wrapper(file_path_or_obj: Optional[Union[str, BinaryIO]]) -> Optional[T]:
...
+#             which loads the native sample rate of the audio file
+#         mono (bool): if True, loads audio as mono, averaging multiple channels
+#             if present. if False, loads audio with the native number of channels.
+
+#     Returns:
+#         * np.ndarray - the mono audio signal
+#         * float - The sample rate of the audio file
+
+#     """
+#     return librosa.load(fhandle, sr=samplerate, mono=mono)
```
drubinstein, line 76: Uh should this remain here?

drubinstein (approved):
Nothing objectionable from a quick glance.

Why is there changes the Maestro code?

mirdata/io.py
rabitt: eek good catch. Will remove!

tests/test_io.py
```diff
@@ -1,11 +1,40 @@
...
+    midi_file = (
+        "tests/resources/mir_datasets/maestro/2018/"
+        + "MIDI-Unprocessed_Chamber3_MID--AUDIO_10_R3_2018_wav--1.midi"
+    )
+    midi = io.load_midi(midi_file)
+    assert len(midi.instruments) == 1
+    assert len(midi.instruments[0].notes) == 4197
+
+
+def test_load_notes_from_midi():
```
nkundiushuti, line 20: do you also test note value?

mirdata/datasets/indexes/slakh_index_baby.json
```diff
@@ -0,0 +1,3717 @@
...
+      "midi": [
+        "babyslakh_16k/Track00001/MIDI/S05.mid",
+        "c40df0d092b1eb74d4c6279ef141c6be"
+      ],
+      "metadata": [
+        "babyslakh_16k/Track00001/metadata.yaml",
+        "ffde21b0625fd72ba04103ca55f6765d"
+      ]
+    },
+    "Track00001-S06": {
```
nkundiushuti, line 88: what's the deal with these effects tracks with null audio and midi? why do they have to be tracks and not metadata? could the record object solve this?

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,332 @@
...
+    @core.cached_property
+    def midi(self) -> Optional[pretty_midi.PrettyMIDI]:
+        return io.load_midi(self.midi_path)
+
+    @core.cached_property
+    def notes(self) -> annotations.NoteData:
+        return io.load_notes_from_midi(self.midi_path, self.midi)
+
+    @property
+    def audio(self) -> Optional[Tuple[np.ndarray, float]]:
```
nkundiushuti, line 173: for some tracks this is null and these tracks are added to the multitrack object. should we check in core.py if the audio is not null?

nkundiushuti (commented):
really great dataset and I am happy you decided to do this loader!

tests/test_io.py
rabitt: what do you mean? In the lines below I test the values, no?

mirdata/datasets/indexes/slakh_index_baby.json
rabitt: Yeah, this was one of my doubts I'm waiting for ethan on. I'm not sure if they're included in the mixture or not - if they're not i'm inclined to remove them as tracks?

mirdata/datasets/slakh.py
rabitt: We already do actually - it's handled in the io.coerce_from_bytes decorator!

ethman:
Hey, sorry for the delay. Right now, I'm playing catchup with a handful of things. I'll get to your questions first:

>There are stems with no midi/audio files. I read in your documentation that you don't render "unusual" sounds. Am I correctly assuming that the midi file for the mixtures also exclude those unusual sources? If that's the case, do you think it's worth including them in the index at all?

Yeah, right. All source audio is guaranteed to have midi, but not all midi has corresponding audio. The mixtures are made by summing the source audio, so those orphaned midi files correspond to no audio anywhere in Slakh. I mostly left them in for completeness. It's up to you if you want to index them, but ignoring them won't harm anything.

> I used slakh2011-redux as the main/default version and baby slakh for tests. I didn't add support for the other two (slakh2100-orig and slakh2100-split2), since it seems like (?) anyone new to the dataset should use redux. Do you agree?

Yup! I'm trying to push people towards redux because of the data leakage in the other two splits (whoops lol). The other two splits are still accessible, but require work to get to. I expect mirdata loader could work for those two splits as well.

> so far I didn't need to use slakh-utils as a dependency, unless we want to support converting between versions. For loading the audio, librosa supports flac and can resample if needed. We support multitrack mixing in core.MultiTrack, but not the recipes you have. Are there certain submixing recipes you think would be useful to add? Let me know if I missed anything.

Most of the stuff in there is to provide a nice on-ramp for people to interface with the dataset, but mirdata already does that, so the majority of stuff in there is OK to omit. For submixing, I think the [five submixes](https://github.com/ethman/slakh-utils/tree/master/submixes/example_submixes) in the repo are probably a good starting place (`bass, drums, guitar, piano, other`, `bass, other`, `drums, other`, `guitar, other`, `piano, other`).

>  If you happen to have a local copy of slakh2100-redux, could you try the block below and let me know if it passes?

Currently running! It'll probably finish after I sign off for today. So far so good, but I will update with the results.

I've done a cursory look over the code and nothing sticks out so far, but I'm still learning about this codebase. I'll circle back for a proper CR in the next few days!

rabitt:
@ethman thanks for the comments! I'll make a few updates now to add the submixing recipes, and to remove tracks from the index which have neither midi nor audio. I'd be happy to have your eyes on the code itself, but if you're swamped, what you've already commented is already great!

mirdata/datasets/indexes/slakh_index_baby.json
nkundiushuti: yes. I think it's better to remove them from the index or from the multi-track list. I am curious what do we do with effects. there are some room impulse responses in phenicx anechoic which were used to build the multi-microphone recordings but I left them out.

mirdata/datasets/slakh.py
nkundiushuti: sorry! you're right!

tests/test_io.py
nkundiushuti: yes! you are right! sorry!

rabitt:
Hey @ethman ! Quick poke to see if you have some time to give this a look.

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+
+class Track(core.Track):
+    """slakh Track class
+
+    Attributes:
+        audio_path (str or None): path to the track's audio file. For some unusual tracks,
+            such as sound effects, there is no audio and this attribute is None.
+        data_split (str or None): one of 'train', 'validation', 'test', or 'omitted'.
+            'omitted' tracks are part of slack2100-redux which were found to be
+            duplicates in the original slackh2011.
```
ethman, line 93: Typo: `original slackh2011` --> `original Slakh2100`. FWIW, it's up to you if you want to capitalize "slakh" or not. I don't think we decided on stylizing it one way or another.

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+            In baby slakh there are no splits, so this attribute is None.
+        metadata_path (str): path to the multitrack's metadata file
+        midi_path (str or None): path to the track's midi file. For some unusual tracks,
+            such as sound effects, there is no midi and this attribute is None.
+        mtrack_id (str): the track's multitrack id
```
ethman, line 98: What's the difference between `mtrack_id` and `track_id`?

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+        track_id (str): track id
+        instrument (str): MIDI instrument class
```
ethman, line 100: Might be nice to link to some of the MIDI documentation, like [the wikipedia page for MIDI](https://en.wikipedia.org/wiki/General_MIDI#Program_change_events).

mirdata/datasets/slakh.py
ethman: Oh, I see how this is structured now. Never mind!

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+        return io.load_midi(self.midi_path)
+
+    @core.cached_property
+    def notes(self) -> Optional[annotations.NoteData]:
+        return io.load_notes_from_midi(self.midi_path, self.midi)
+
+    @core.cached_property
+    def multif0(self) -> Optional[annotations.MultiF0Data]:
+        return io.load_multif0_from_midi(
+            self.midi_path, self.midi, skip_drums=True, pitch_bend=False
```
ethman, line 298: I believe some (if not all) of the Kontakt patches used to make slakh honor pitch bend events. There are definitely guitar pitch bends in slakh.

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+MIXING_GROUPS = {
+    "piano": [0, 1, 2, 3, 4, 5, 6, 7],
+    "guitar": [24, 25, 26, 27, 28, 29, 30, 31],
+    "bass": [32, 33, 34, 35, 36, 37, 38, 39],
+    "drums": [128],
+}
+
+
+class Track(core.Track):
+    """slakh Track class
```
ethman, line 86: Maybe mention that it represents a "stem" or isolated instrument recording?

mirdata/datasets/slakh.py
```diff
@@ -0,0 +1,408 @@
...
+    def to_jams(self):
+        """Jams: the track's data in jams format"""
+        return jams_utils.jams_converter(
+            audio_path=self.audio_path,
+            note_data=[(self.notes, "Notes")],
+        )
+
+
+class MultiTrack(core.MultiTrack):
+    """slakh multitrack class
```
ethman, line 215: Maybe mention that this has the mix and a set of "stems"?

ethman (approved):
Apologies for the delay and thank you for pinging me on this again!!!! I just dug into this, and everything looks good for the most part. I had a few minor comments and they were mostly about documentation where I was confused.

Overall: LGTM. Thanks for doing this! :-)

mirdata/datasets/slakh.py
rabitt: hmmm is there a way to know for which ones the pitch bends are honors and the ones where it's not? I spot checked a few random ones and they weren't in the audio.

rabitt: (For now i'm going to make a comment and open an issue, and leave it for the future)

ethman: Unfortunately there's no way to know from looking only at the MIDI. Though I think we could get a rough idea of which patches honor pitch bends by finding a few examples of each patch and inspecting the audio by hand. But to be 100% certain, we'd probably need a way to systematically check _all_ of the pitch bend instances using Label Studio or something similar 😫.

PR #499, opened 2021-04-22
Billboard audio is not available on request

chrisdonahue:
A one-liner indicating that, unfortunately, the Billboard audio is not available on request. I reached out to Ashley Burgoyne (the dataset's creator) who informed me that the audio could not be shared. CC @Spijkervet who originally added the dataset to this library, on the off chance that there is some other path for obtaining the audio.


rabitt (approved):
Thanks for this @chrisdonahue !

PR #504, opened 2021-04-28
Annotation Extensions

### LINKED ISSUE #470: Expanding NoteData and F0Data classes
Hello,

I am working on the implementation of the [TONAS dataset](https://zenodo.org/record/1290722). It has f0 and notes annotations, and both annotations have **energy** information. I was thinking about expanding `F0Data` and `NoteData` classes in the loader into `F0DataTonas` and `NoteDataTonas` to be able to consider the new attribute energy.
However, I can also expand the annotations themselves in `annotations.py`, and then we would leave the `energy` attribute empty for the datasets that do not have energy annotations. 

@rabitt, @magdalenafuentes, @nkundiushuti, what are your thoughts about that?

### LINKED ISSUE #501: Annotation unit support
In annotations.py, our annotation classes have fixed units (defined in the docstring), which are not strictly enforced. 

I'm working on a PR to
* allow flexible units/schemas
* enforce that annotations match the given units/schemas

This will close #470

rabitt:
* Adds units to all Annotation classes, which will close #501
* Separates voicing and confidence in F0Data classes, with an API change warning
* supports different units of confidence, which will close #470 
* (a few bonus changes) as I was updating the loaders, I fixed some inconsistences between them (some weren't using the file io decorators, removed unused imports, etc)


rabitt:
cc @PRamoneda @genisplaja @nkundiushuti - If you have time to give a review, I'd be happy to have it, but no problem if you're busy!

tests/test_loaders.py
```diff
@@ -333,11 +333,13 @@ def test_track_placeholder_case():
     "guitarset": {
         "load_pitch_contour": {"string_num": 1},
         "load_notes": {"string_num": 1},
+        "load_chords": {"leadsheet_version": False},
     },
     "saraga": {
         "load_tempo": {"iam_style": "carnatic"},
         "load_sections": {"iam_style": "carnatic"},
```
genisplaja, line 340: This is outdated, we could take the chance to remove it in this PR!

genisplaja:
Hey @rabitt! I briefly took a look at the changes, and they look really good to me:) This is great!! Maybe I missed it but, there is no information about the introduction of `units` in the contributing documentation right? I don't know if you're waiting for more reviews before expanding the contributing documentation, but I just wanted to point out that it would be really good to talk about the units there. Otherwise, we should at least update the `example.py` in contributing examples!
Thanks!!

tests/test_loaders.py
genisplaja: When I split `saraga` in the two collections I forgot to remove that... sorry!

rabitt:
> Hey @rabitt! I briefly took a look at the changes, and they look really good to me:) This is great!! Maybe I missed it but, there is no information about the introduction of units in the contributing documentation right? I don't know if you're waiting for more reviews before expanding the contributing documentation, but I just wanted to point out that it would be really good to talk about the units there. Otherwise, we should at least update the example.py in contributing examples!
Thanks!!

Whoops yes good point!! I totally forgot to update the contributing docs. I'll do that now. Thanks 🙏

mirdata/annotations.py
```diff
@@ -111,143 +237,242 @@ class F0Data(Annotation):
...
+    def value(self):
+        logging.warning(
+            "Deprecation warning: TempoData.value will be removed in future versions. Use TempoData.tempos instead."
+        )
+        return self.tempos
 
 
 class EventData(Annotation):
-    """TempoData class
+    """EventData class
```
magdalenafuentes, line 451: nice catch

mirdata/annotations.py
```diff
@@ -319,38 +544,239 @@ def validate_lengths_equal(array_list):
...
+
+    if position_unit in ["bar_index", "global_index"] and not np.array_equal(
+        np.floor(positions), positions
+    ):
+        raise ValueError(
+            "measure index or global indexes should be integers. "
+            + "Found fractional values."
+        )
+
+    if position_unit == "bar_index" and np.max(positions) > 16:
```
magdalenafuentes, line 592: Why 16?

mirdata/annotations.py
```diff
@@ -319,38 +544,239 @@ def validate_lengths_equal(array_list):
...
         raise ValueError(
-            f"Confidence should be 1d, but array has shape {confidence_shape}"
+            "confidence with unit 'likelihood' should be between 0 and 1. "
+            + "Found values outside [0, 1]."
         )
 
-    if (confidence < 0).any() or (confidence > 1).any():
-        raise ValueError("confidence should be between 0 and 1")
+    if confidence_unit == "energy" and any([c < 0 for c in confidence_flat]):
+        raise ValueError(
```
magdalenafuentes, line 637: Line not covered by tests

rabitt:
Thanks @magdalenafuentes !

> My feeling is that we’re converging more and more to the style of JAMS, and standardizing the datasets. I guess is a good thing, but I’m a bit concern whether all datasets comply with the formats you’re including, and we can’t see that in the PR. 

Yes I agree, and it's the direction I was trying to move in - eventually using jams annotations under the hood, keeping the api the same. I'd argue(?) that this PR is actually allowing more flexibility for different data types, rather than standardizing - before we forced one particular "unit", and sometimes we had annotations that silently didn't match what was expected (e.g. notes with midi frequencies instead of hz). 

> It would require to run all test_full_dataset right?

Good point... It is possible some datasets will now fail when some annotations are initialized because of the annotation validation. Would you be opposed to letting those potential bugs come up as they're discovered? There shouldn't be any new bugs, there's just now code that will catch them if they're there. 

> I see you have all units mandatory right now, I guess you want to push contributors to reflect on the correctness of their annotations and I agree with that. I have a minor concern about adding yet more decision making in the process of adding a loader.

I was thinking less about correctness, and more about preciseness. We already ask to have units in the docstrings, so the contributors will probably already know what the units are. what do you think?

docs/source/contributing.rst
```diff
@@ -361,14 +361,6 @@ We've added one extra convenience flag for this test, for getting the tests runn
 
 which will skip the downloading step. Note that this is just for convenience during debugging - the tests should eventually all pass without this flag.
 
-.. _working_big_datasets:
```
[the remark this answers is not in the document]
rabitt: there's a new section about large datasets and large indexes, and it's covered there!

mirdata/annotations.py
```diff
@@ -16,91 +84,149 @@ class BeatData(Annotation):
...
+        confidence (np.ndarray): array of confidence values
+        confidence_unit (str): confidence unit, one of CONFIDENCE_UNITS
 
     """
 
-    def __init__(self, times, positions=None):
+    def __init__(
+        self,
+        times,
+        time_unit,
```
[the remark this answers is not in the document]
rabitt: I wanted to have it not default on purpose so that when the loader is defined, we have to think about what the units should be. (to avoid very evil secret bugs when we use them!)

mirdata/annotations.py
rabitt: I was thinking 16 was the max common number of beats per bar. Maybe I'm wrong? :)

magdalenafuentes: I think this works for the datasets we have, but if later one some `MTG`'s Indian music datasets are added we might see some issues (IIRC those can have >20 beats in a cycle). We can change this number then  :)

magdalenafuentes:
I think it would be ok doing a `rc` release with a warning, and I think problems should arise in the weekly tests? ccing @nkundiushuti 

I'm fine in asking more precision in the annotations :)

mirdata/annotations.py
rabitt: Ah ok! I'll update it now! it was mostly to catch the difference between bar_index and global_index.

rabitt:
> I think it would be ok doing a rc release with a warning, and I think problems should arise in the weekly tests? 

An rc release sounds good to me!

magdalenafuentes (approved):

PR #507, opened 2021-05-01
Add note annotations for medleydb_pitch and ikala

rabitt:
For both medleydb_pitch and ikala:
* Adds note annotations (estimated by pyin)
* updates index scripts
* adds new versions of indexes

I ran test_full_dataset for both datasets.


drubinstein (commented):
LGTM

PR #513, opened 2021-05-09
speed up deduplication in NoteData

rabitt:
`NoteData._remove_duplicates` is new as of this beta version, and it runs every time a NoteData object is created. but its very slow... this speeds it up.


rabitt:
> np.unique returns the sorted unique elements. Can this be a problem? If not then I approve.

It's actually a good thing for this that we weren't properly handling before. Before note events could be in any order, now they'll be sorted by start time.

PR #515, opened 2021-06-03
Add MultiAnnotator class

magdalenafuentes:
This PR adds the MultiAnnotator class to group annotations of same track by different annotators in a smooth and clear way.

### DIFF
--- a/mirdata/annotations.py
+++ b/mirdata/annotations.py
@@ -79,6 +79,26 @@ class Annotation(object):
         return repr_str
 
 
+class MultiAnnotator(Annotation):
+    """Multiple annotator class.
+    This class should be used for datasets with multiple annotators (e.g. multiple annotators per track).
+
+    Attributes:
+        annotators (list): list with annotator ids
+        annotations (list): list of annotations (e.g. [annotations.BeatData, annotations.ChordData]
+    """
+
+    def __init__(self, annotators, annotations) -> None:
+        validate_array_like(annotators, list, str, none_allowed=True)
+        validate_array_like(
+            annotations, list, Annotation, check_child=True, none_allowed=True
+        )
+        validate_lengths_equal([annotators, annotations])
+
+        self.annotators = annotators
+        self.annotations = annotations
+
+
 class BeatData(Annotation):
     """BeatData class
 
@@ -1383,7 +1403,9 @@ def closest_index(input_array, target_array):
     return indexes
 
 
-def validate_array_like(array_like, expected_type, expected_dtype, none_allowed=False):
+def validate_array_like(
+    array_like, expected_type, expected_dtype, check_child=False, none_allowed=False
+):
     """Validate that array-like object is well formed
 
     If array_like is None, validation passes automatically.
@@ -1392,11 +1414,12 @@ def validate_array_like(array_like, expected_type, expected_dtype, none_allowed=
         array_like (array-like): object to validate
         expected_type (type): expected type, either list or np.ndarray
         expected_dtype (type): expected dtype
+        check_child (bool): if True, checks if all elements of array are children of expected_dtype
         none_allowed (bool): if True, allows array to be None
 
     Raises:
         TypeError: if type/dtype does not match expected_type/expected_dtype
-        ValueError: if array
+        ValueError: if array is empty but it shouldn't be
 
     """
     if array_like is None:
@@ -1416,8 +1439,20 @@ def validate_array_like(array_like, expected_type, expected_dtype, none_allowed=
         )
 
     if expected_type == list and not all(
-        isinstance(n, expected_dtype) for n in array_like
+        isinstance(n, expected_dtype)
+        for n in array_like
+        if not ((n is None) and none_allowed)
     ):
+        if check_child:
+            if not all(
+                issubclass(type(n), expected_dtype)
+                for n in array_like
+                if not ((n is None) and none_allowed)
+            ):
+                raise TypeError(
+                    f"List elements should all be instances of {expected_dtype} class"
+                )
+
         raise TypeError(f"List elements should all have type {expected_dtype}")
 
     if (
--- a/mirdata/datasets/salami.py
+++ b/mirdata/datasets/salami.py
@@ -158,20 +158,30 @@ class Track(core.Track):
         return self._track_metadata.get("genre")
 
     @core.cached_property
-    def sections_annotator_1_uppercase(self) -> Optional[annotations.SectionData]:
-        return load_sections(self.sections_annotator1_uppercase_path)
+    def sections_uppercase(self) -> Optional[annotations.MultiAnnotator]:
+        return annotations.MultiAnnotator(
+            [
+                self._track_metadata.get("annotator_1_id"),
+                self._track_metadata.get("annotator_2_id"),
+            ],
+            [
+                load_sections(self.sections_annotator1_uppercase_path),
+                load_sections(self.sections_annotator2_uppercase_path),
+            ],
+        )
 
     @core.cached_property
-    def sections_annotator_1_lowercase(self) -> Optional[annotations.SectionData]:
-        return load_sections(self.sections_annotator1_lowercase_path)
-
-    @core.cached_property
-    def sections_annotator_2_uppercase(self) -> Optional[annotations.SectionData]:
-        return load_sections(self.sections_annotator2_uppercase_path)
-
-    @core.cached_property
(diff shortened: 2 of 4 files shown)


mirdata/annotations.py
```diff
@@ -1392,11 +1414,12 @@ def validate_array_like(array_like, expected_type, expected_dtype, none_allowed=
         array_like (array-like): object to validate
         expected_type (type): expected type, either list or np.ndarray
         expected_dtype (type): expected dtype
+        check_child (bool): if True, checks if all elements of array are children of expected_dtype
```
rabitt, line 1417: what happens if there are multiple children with different expected dtypes (e.g a beat annotation with ints for the beat positions and floats for the time stamps?)

mirdata/datasets/salami.py
```diff
@@ -158,20 +158,30 @@ def genre(self):
...
-        return load_sections(self.sections_annotator1_uppercase_path)
-
-    @core.cached_property
-    def sections_annotator_1_lowercase(self) -> Optional[annotations.SectionData]:
-        return load_sections(self.sections_annotator1_lowercase_path)
-
-    @core.cached_property
-    def sections_annotator_2_uppercase(self) -> Optional[annotations.SectionData]:
-        return load_sections(self.sections_annotator2_uppercase_path)
+    def sections_uppercase(self) -> Optional[annotations.MultiAnnotator]:
```
rabitt, line 161: to make this a non-breaking change, we can deprecate it by keeping the existing attributes `sections_anntotator_1_uppercase` as they are but logging a deprecation warning, and pointing to the new attribute they should use:

something like:
```def sections_anntotator_1_uppercase(self):
    logging.warning("Deprecation warning: sections_anntotator_1_uppercase is deprecated starting in version <whatever version we are now> and will be removed in a future version. Use sections_uppercase in the future.
```

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
         return repr_str
 
 
+class MultiAnnotator(Annotation):
+    """Multiple annotator class.
+    This class should be used for datasets with multiple annotators (e.g. multiple annotators per track).
+
+    Attributes:
+        annotators (list): list with annotator ids
+        annotations (list): list of annotations (e.g. [annotations.BeatData, annotations.ChordData]
```
nkundiushuti, line 88: as far as I see in salami, the list of annotations comprises lists of annotations rather than annotations
it's not clear what happens when there are multiple annotation types and multiple annotators 
for instance, is this the structure for two annotators:
[[annotations.BeatData,annotations.BeatData], [annotations.ChordData,annotations.ChordData]]
or this
[[annotations.BeatData,annotations.ChordData], [annotations.BeatData,annotations.ChordData]]

nkundiushuti (changes requested):
maybe add something to the contributing documentation?

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
         return repr_str
 
 
+class MultiAnnotator(Annotation):
```
rabitt, line 82: nit: does it make sense for this to be a subclass of `Annotation`? I think it might be better to just subclass `object`

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
...
+
+    def __init__(self, annotators, annotations) -> None:
+        validate_array_like(annotators, list, str, none_allowed=True)
+        validate_array_like(
+            annotations, list, Annotation, check_child=True, none_allowed=True
+        )
+        validate_lengths_equal([annotators, annotations])
+
+        self.annotators = annotators
+        self.annotations = annotations
```
rabitt, line 99: maybe too much for this PR, but thought for the future - 

this class could have methods like: 
`.get(annotator_id)`
`.get_best()`  (which uses a dataset-defined heirarchy when initialized saying which annotations are "best")

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
...
+class MultiAnnotator(Annotation):
+    """Multiple annotator class.
+    This class should be used for datasets with multiple annotators (e.g. multiple annotators per track).
+
+    Attributes:
+        annotators (list): list with annotator ids
+        annotations (list): list of annotations (e.g. [annotations.BeatData, annotations.ChordData]
+    """
+
+    def __init__(self, annotators, annotations) -> None:
```
rabitt, line 91: I think these two inputs would be better as a single dictionary:
{ 'annotator-id': Annotation }

same with how it's stored, self.annotations = `{'annotator1': ... , 'annotator2', ...}`
because it will make it easier to look up a specific dictionary instead of having to look for matching indexes.

rabitt (changes requested):

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
...
+        validate_array_like(annotators, list, str, none_allowed=True)
+        validate_array_like(
+            annotations, list, Annotation, check_child=True, none_allowed=True
```
rabitt, line 94: should the validation also check that the `Annotation` subtypes are all the same?

mirdata/annotations.py
magdalenafuentes: here we're only checking if the list is composed of `BeatData`, `ChordData`, etc. The checks within the annotation type are done for each annotation type independently

mirdata/annotations.py
magdalenafuentes: in that case, you would have `tracks.chords` and `tracks.beats` and each would return a `MultiAnnotator` class with format `MultiAnnotator(annotations, annotators)`,  e.g. `annotations` being [`BeatData`, `BeatData`] and `annotators` ["annot1", "annot2"]

mirdata/datasets/salami.py
magdalenafuentes: It's true, though I suspect that the `salami` loader doesn't have many users since the audio is private XD, so I'm fine with the breaking change actually. Thoughts?

mirdata/annotations.py
magdalenafuentes: But semantically it is an annotation no? Plus it prints so nicely <3

mirdata/annotations.py
magdalenafuentes: Yeah, plus we could have methods with different ways of combining the annotations, which in some cases make sense! Like minimum vote, or all annotators agree, etc. We can do that in a follow-up PR

mirdata/annotations.py
rabitt: I think the issue is that it's ambiguous from the way the docs are written. 

proposal:
```
list of annotations (e.g. [beat_data1, beat_data2] each with type BeatData or  [chord_data1, chord_data2] each with type chord data
```

mirdata/datasets/salami.py
rabitt: so far in the other loaders I've been doing the deprecation thing, and I think it's good practice. You don't have to change any of your current code, just copy back the existing attributes and add the deprecation warnings.

mirdata/annotations.py
rabitt: we might want to print multi-annotations differently than Annotations though, no?

mirdata/annotations.py
```diff
@@ -79,6 +79,26 @@ def __repr__(self):
         return repr_str
 
 
+class MultiAnnotator(Annotation):
+    """Multiple annotator class.
+    This class should be used for datasets with multiple annotators (e.g. multiple annotators per track).
+
+    Attributes:
```
rabitt, line 86: as I'm thinkign about this more, I think the annotation type should be passed as an argument to init, and used in the validation. In the future I think it will save us a lot of code.

mirdata/annotations.py
magdalenafuentes: I tried this out and noticed that some tracks have the same `annotator id` for different annotations in the same track (which might be a bit weird actually... but it's there :/). Also the sintax gets a bit messy, since I have to iterate in the keys each time I want to access an annotation  because the annotator id changes from track to track. So I changed back to what was before for now. Let me know what you think

nkundiushuti (approved):
this looks great! thanks, @magdalenafuentes !

magdalenafuentes:
Will wait for @rabitt to see if she has any other comments/edits. Otherwise @nkundiushuti or @rabitt feel free to merge since I will be offline starting today!

Issue #462, opened 2021-02-03
Support remote filepaths


rabitt:
When data is in a remote storage location e.g. AWS, Google Cloud Storage, Google Drive, etc, using mirdata becomes clunky, and we've overloaded the API to support it. Instead, it would be great to simplify the API and instead directly support remote file paths.

I propose we start by directly supporting Google Cloud Storage filepaths (`gs://...`) so they work seamlessly.

cc @drubinstein @psobot

drubinstein:
I forget what the current interface looks like, but maybe it'd be better if the user is responsible for giving a callback that handles finding the file and downloading it (or giving it to a Bytes/StringIO) and mirdata handles the deserialization. There will always be another way to download a file.

rabitt:
> I forget what the current interface looks like, but maybe it'd be better if the user is responsible for giving a callback that handles finding the file and downloading it (or giving it to a Bytes/StringIO) and mirdata handles the deserialization. There will always be another way to download a file.

drubinstein:
You could write a bunch of callbacks and the user specifies which one they want to use. That way they can use the file one or maybe a GCS one if you want to add it. This way a user could add in their own downloader if need be. I think my thought is instead of focusing on supporting "all remote filepaths" it may be more useful to first focus on making the interface extensible.


A "universal" downloader sounds like an independent project and maybe could be another python package in this organization that this repo uses.

drubinstein:
An alternative to writing an in-house library is to use [cloudstorage](https://pypi.org/project/cloudstorage/) which seems to fit this use case.

PR #518, opened 2021-08-05
Use smart_open

rabitt:
Work in progress to address #462 . This change so far seems to work equivalently for local filesystems - tbd if it works for remote filesystems.

* adds a new dependency, [smart_open](https://github.com/RaRe-Technologies/smart_open)
* replaces calls to `open` in core code, with some modifications for compatibility (namely, removes most calls to os.path.exists)
* raises `NotImplementedError` if a user tries to download to a remote filesystem.

Still todo:
- [x] update individual loaders which use `open`
- [x] add tests for remote filesystems
- [x] add examples using remote filepaths in docs & deprecate existing docs

### DIFF
--- a/mirdata/core.py
+++ b/mirdata/core.py
@@ -177,12 +177,12 @@ class Dataset(object):
         try:
             with open(self.index_path) as fhandle:
                 index = json.load(fhandle)
-        except IOError:
+        except FileNotFoundError:
             if self._index_data.remote:
                 raise FileNotFoundError(
                     "This dataset's index must be downloaded. Did you run .download()?"
                 )
-            raise IOError(
+            raise FileNotFoundError(
                 f"Dataset index for {self.name} was expected "
                 + "to be packaged with mirdata, but not found."
             )
--- a/mirdata/core.py
+++ b/mirdata/core.py
@@ -6,8 +6,8 @@ import random
 import types
 from typing import Any, List, Optional
 
-
 import numpy as np
+from smart_open import open
 
 from mirdata import download_utils
 from mirdata import validate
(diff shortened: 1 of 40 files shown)


mirdata/core.py
```diff
@@ -174,14 +174,18 @@ def __repr__(self):
...
+            with open(self.index_path) as fhandle:
+                index = json.load(fhandle)
+        except IOError:
+            if self._index_data.remote:
+                raise FileNotFoundError(
+                    "This dataset's index must be downloaded. Did you run .download()?"
+                )
             raise IOError(
-                "Dataset index was expected to be availale locally, but not found."
+                "Dataset index was expected to be packaged with mirdata, but not found."
```
drubinstein, line 186: Would add the index's name

mirdata/validate.py
```diff
@@ -48,7 +50,9 @@ def validate(local_path, checksum):
 
     """
     # validate that the file exists on disk
-    if not os.path.exists(local_path):
+    try:
+        open(local_path)
```
drubinstein, line 54: shouldn't this be a with open with a pass? Otherwise I belive you still need to cache the variable and close it

tests/test_utils.py
```diff
@@ -91,15 +91,16 @@ def mock_validate_index(mocker):
     return mocker.patch.object(validate, "validate_index")
 
 
-def test_md5(mocker):
-    audio_file = b"audio1234"
+# TODO: rewrite this!
```
drubinstein, line 94: Can you explain how or why to rewrite this?

mirdata/io.py
```diff
@@ -4,6 +4,7 @@
 
 import numpy as np
 import pretty_midi
+from smart_open import open
```
drubinstein, line 7: I think you can delete this line

rabitt: I think it'd needed - we call the system `open` in the decorator, and we'll need to overwrite it with smart_open to support remote files, no?

tests/test_utils.py
rabitt: the test fails now, and we're patching builtings.open, instead of smart_open

drubinstein: Thanks. Can you add that to the comment?

mirdata/io.py
drubinstein: Ah forgot that was in this file. nvm then.

mirdata/datasets/guitarset.py
```diff
@@ -490,9 +496,12 @@ def load_notes(jams_path, string_num):
         NoteData: Note data for the given string
 
     """
-    if not os.path.exists(jams_path):
-        raise IOError("jams_path {} does not exist".format(jams_path))
-    jam = jams.load(jams_path)
+    try:
```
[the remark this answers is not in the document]
rabitt: there's a comment about it on line 486!

drubinstein (approved):

Issue #495, opened 2021-04-14
Option to ignore md5 checksum inconsistencies during download


rabitt:
In some exceptional cases, the md5 checksum of a downloaded file may not match what we expect (e.g. see #494 ), but after decompression, the internal files have the expected checksums.

To avoid this, we could add a `allow_invalid_checksum` boolean option to `download_utils.downloader` (which defaults to `False`), which prints a warning when a checksum is invalid but does not raise an exception, and prompts the user to run validation afterward to ensure the integrity of the downloaded data.

(Thanks @migperfer for pointing out this issue and for the idea!)

magdalenafuentes:
I like this idea! 

What if we make `cleanup` dependent on that boolean, and if we're not checking `checksums` by default when we download we delete the compressed files?

migperfer:
I guess this one is done then? :)

nkundiushuti:
solved in #530

PR #524, opened 2021-09-23
Ignore checksum

migperfer:
Hi! Sorry for the delay. I added the option to ignore the checksum (see issue #495) and updated the tests with them. Should I do something as well with the documentation, or is it automatically generated?

### DIFF
--- a/mirdata/download_utils.py
+++ b/mirdata/download_utils.py
@@ -123,11 +123,25 @@ def downloader(
             logging.info("[{}] downloading {}".format(k, remotes[k].filename))
             extension = os.path.splitext(remotes[k].filename)[-1]
             if ".zip" in extension:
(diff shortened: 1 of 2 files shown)


rabitt:
This looks great @migperfer, thanks ! It looks good to go, except for the formatting check. A quick fix is to run:
```
tox -e format
```
from the top-level repository folder and commit the changes. If that doesn't work, you can instead run:
```
black --target-version py38 mirdata tests
```
Let me know if you run into any problems.

migperfer:
Hi! Sorry for the mess I made with the commits and merges. To what is the `codecov/patch` error?

migperfer:
For some reason on my screen appears that buildpy38/37 passed without problems, but when I was talking with @genisplaja on his screen they appear as failed. It could be caused by the messy git management that I did with my local copy, I will just close this pull request and open another one without the mess, lets see what happens.

PR #527, opened 2021-09-27
pin music 21 version

rabitt:
The latest builds are all failing because music21 updated in an incompatible way and we hand't pinned the version. This pins it!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,6 +62,6 @@ if __name__ == "__main__":
                 "sphinx_rtd_theme",
             ],
             "dali": ["dali-dataset==1.1"],
-            "haydn_op20": ["music21"],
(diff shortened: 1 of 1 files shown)


genisplaja (approved):

setup.py
```diff
@@ -62,6 +62,6 @@
                 "sphinx_rtd_theme",
             ],
             "dali": ["dali-dataset==1.1"],
-            "haydn_op20": ["music21"],
+            "haydn_op20": ["music21==6.7.1"],
```
drubinstein, line 65: Generally bad idea in a library to pin exactly. Maybe use <=6.7.1?

rabitt: the latest version is what's breaking it :(

Issue #516, opened 2021-06-11
Datasets with train/test/validation splits


migperfer:
Hi! I was thinking about adding [Magenta's NSynth dataset](https://magenta.tensorflow.org/datasets/nsynth). The problem is that it has 3 different splits (train, test, validation), and I'm not sure about your thoughts on _mirdata_ handling this. Could it be possible to add datasets with splits to _mirdata_, or is it better to just add the three of them together?

genisplaja:
I would suggest to add the three splits in a single loader. You can add an attribute to the class `Track` called, for instance, `split`, which is a string that can be `train`, `validation` or `test`. Thus, users will be able to filter the tracks by split by just checking what `Track.split` is.
See IRMAS loader for an example of that.

migperfer:
Thanks for the idea and the example!

But this situation gave me an idea. Wouldn´t be nice to actually have some sort of splitting tool inside _mirdata_ to enhance reproducibility? For example, let's say that I split `dataset A`  into `train` and `test` with a fixed `seed` so other people can reproduce my results.

```
orchset = mirdata.initialize('orchset')  # get the orchset dataset
orchset.download()  # download orchset
orchset.validate()  # validate orchset
tracks_train, tracks_validation = orchset.split_dataset([0.7, 0.3], seed=0.40)  # easily reproduce train/validation splits
...
```

rabitt:
> But this situation gave me an idea. Wouldn´t be nice to actually have some sort of splitting tool inside mirdata to enhance reproducibility? For example, let's say that I split dataset A into train and test with a fixed seed so other people can reproduce my results.

I agree it's a good idea! It could be part of the `Dataset` base clase for example.

Laubeee:
@migperfer Did you complete the NSynth data loder? any chance to add it to the main repo?

migperfer:
I actually did not. I made the splitting functions, but I completely forgot about the NSynth dataset unfortunately 🫠

PR #530, opened 2021-10-06
Dataset splits & Checksum inconsistencies

migperfer:
I forked again the repo and added the proposed functions in issues #516 and #495.


mirdata/core.py
```diff
@@ -283,6 +283,58 @@ def choice_multitrack(self):
...
+
+        if not np.isclose(np.sum(splits), 1):
+            raise ValueError(
+                "Splits values should sum up to 1. Given {} sums {}".format(
+                    splits, np.sum(splits)
+                )
+            )
+
+        rng = np.random.default_rng(seed=seed)  # The random generator
+        random_track_ids = rng.permutation(self.track_ids)
```
magdalenafuentes, line 305: Maybe you can add a condition to check that the dataset has tracks? and the same for multitracks? to avoid triggering an error hard to debug in this line or on the code below?

mirdata/download_utils.py
```diff
@@ -235,16 +246,29 @@ def download_from_remote(remote, save_dir, force_overwrite):
...
+                UserWarning,
+            )
+        else:
+            raise IOError(
+                "{} has an MD5 checksum ({}) "
+                "differing from expected ({}), "
+                "file may be corrupted.".format(
+                    download_path, checksum, remote.checksum
+                )
+            )
```
magdalenafuentes, line 265: :heart:

tests/test_download_utils.py
```diff
@@ -42,63 +43,63 @@ def test_downloader(mocker, mock_path):
     )
 
     zip_remote = download_utils.RemoteFileMetadata(
-        filename="remote.zip", url="a", checksum="1234"
+        filename="remote.zip", url="a", checksum=("1234")
```
magdalenafuentes, line 46: Sorry but I'm missing why you need the `()` here, could you explain?

tests/test_loaders.py
```diff
@@ -466,3 +466,46 @@ def test_multitracks():
...
+def test_random_splits():
+    for dataset_name in DATASETS:
+        dataset = mirdata.initialize(
+            dataset_name, os.path.join(TEST_DATA_HOME, dataset_name), version="test"
+        )
+
+        list_sum_up_1 = [
+            [0.8, 0.1, 0.1],
+            [0.7, 0.2, 0.1],
+            [0.9, 0.1, 0.0],
```
magdalenafuentes, line 480: Maybe you can add a couple of examples with different lengths to see if it does what you expect?

magdalenafuentes: Basically you could add a small test checking that you receive the number of splits that you expect

magdalenafuentes (commented):
Hey @migperfer thanks a lot for this, it's great! And sorry for the slow turnaround. I just left two minor comments and a question, other than that is great and good to merge!

tests/test_download_utils.py
migperfer: I do not remember doing this 😯 I could think that it was done by mistake, but it has been done consistently across all the code (see lines 49, 53, 164, 167, etc) Maybe a black/tox thing?

mirdata/core.py
migperfer: Done! Unfortunately, my computer is kinda Frankenstein and I can not run the tests on my local setup. I will try to find some alternative where I can run the write & run tests and meet the coverage expectations.

tests/test_loaders.py
migperfer: Done! 👍

tests/test_loaders.py
```diff
@@ -466,3 +466,66 @@ def test_multitracks():
...
+                    dataset.get_tracks_splits(wrong_combination)
+
+        # check splits for multitracks
+        if dataset._multitrack_class:
+            for right_combination in list_sum_up_1:
+                splits = dataset.get_mtracks_splits(right_combination)
+                assert len(dataset.mtrack_ids) == sum([len(i) for i in splits])
+            for wrong_combination in list_not_sum_up_1:
+                with pytest.raises(ValueError):
+                    dataset.get_mtracks_splits(wrong_combination)
```
rabitt, line 531:
```suggestion
def test_random_splits():
    split = [0.9, 0.1]
    for dataset_name in DATASETS:
        dataset = mirdata.initialize(
            dataset_name, os.path.join(TEST_DATA_HOME, dataset_name), version="test"
        )
        
        # check wrong type of split function
        if dataset._track_class is None:
            with pytest.raises(AttributeError):
                dataset.get_track_splits(split)

        if dataset._multitrack_class is None:
            with pytest.raises(AttributeError):
                dataset.get_mtrack_splits(split)

        # check splits for tracks
        if dataset._track_class:
            splits = dataset.get_tracks_splits(split)
            assert len(dataset.track_ids) == sum([len(i) for i in splits])

        # check splits for multitracks
        if dataset._multitrack_class:
                splits = dataset.get_mtracks_splits(split)
                assert len(dataset.mtrack_ids) == sum([len(i) for i in splits])

```

This is a great test! I think it's actually doing much more than we need so here's my suggestion to simplify this test, and a suggested test to add to `test_core`. There's no way in github for me to directly suggest this, so here's my suggested function to add to `tests/test_core.py`:

```python
def test_dataset_splits():
    empty_dataset = core.Dataset()
    
    # test the case where there are no tracks
    with pytest.raises(AttributeError):
        empty_dataset.get_track_splits([0.9, 0.1])
    
    # test the case where there are no multitracks
    with pytest.raises(AttributeError):
        empty_dataset.get_mtrack_splits([0.9, 0.1])

    # test the partition function
    items = [i for i in range(100)]
    list_sum_up_1 = [
        [0.7, 0.1, 0.1, 0.1],
        [0.4, 0.2, 0.2, 0.2],
        [0.5, 0.2, 0.2, 0.1],
        [0.8, 0.1, 0.1],
        [0.7, 0.2, 0.1],
        [0.9, 0.1, 0.0],
        [0.6, 0.3, 0.1],
        [0.5, 0.4, 0.1],
        [0.9, 0.05, 0.05],
        [0.8, 0.2],
        [0.1, 0.9],
    ]
    for right_combination in list_sum_up_1:
        splits = empty_dataset._get_partitions(items, right_combination, 42)
        # check that the right number of splits are created
        assert len(splits) == len(right_combination)
        # check that the number of total items matches
        assert len(items) == sum([len(i) for i in splits])
        # check that all items are used
        assert set(items) == set([i for split_items in splits for i in split_items])
        # check that splits are nonoverlapping
        used = set()
        for split in splits:
            this_split = set(split)
            assert not this_split.intersection(used)
            used.update(this_split)
        # check that the split is reproducable
        splits2 = empty_dataset._get_partitions(items, right_combination, 42)
        assert splits == splits2
        
    list_not_sum_up_1 = [
        [0.8, 0.1, 0.3, 0.2],
        [0.3, 0.1, 0.3, 0.5],
        [0.8, 0.1, 0.3],
        [0.3, 0.1, 0.3],
        [0.9, 0.2, 0.3],
        [0.1, 0.1, 0.1],
        [0.95, 0.01, 0.01],
        [0.8, 0.1, 0.3],
        [0.8, 0.3],
        [0.1, 0.7],
    ]
    for wrong_combination in list_not_sum_up_1:
        with pytest.raises(ValueError):
            empty_dataset._get_partitions(items, wrong_combination, 42)

```

rabitt (commented):
Hey @migperfer ! Thanks for this PR, and sorry for our collective delay in reviewing.

This looks great - I had some ideas for how to reduce the code a bit, and I've included them as suggested changes (you should be able to just "add to batch" and commit them directly in the PR). You might have to rerun black after. Note that there is one change I couldn't directly suggest, which is to split up your existing test for the splitting function, and add one test (code in the comment) to test_core.py.

Let me know if you have any questions!

mirdata/download_utils.py
```diff
@@ -181,14 +200,6 @@ def download_from_remote(remote, save_dir, force_overwrite):
...
 
     """
-    file_uri = parse_uri(save_dir)
-    if file_uri.scheme != "file":
-        raise NotImplementedError(
-            "mirdata only supports downloading to a local filesystem. "
-            "To use mirdata with a remote filesystem, download to a local filesytem, "
-            "and transfer the data to your remote filesystem, setting data_home appropriately."
-        )
-
```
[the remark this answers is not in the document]
migperfer: Mmmm true! Why did not the tests return an error though?

tests/test_loaders.py
migperfer: Great! I will add this test to `test_core` during this week.

migperfer: The line `empty_dataset = core.Dataset()` is giving me problems, any suggestions?

```python
>       empty_dataset = core.Dataset()

tests/test_core.py:729: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
mirdata/core.py:116: in __init__
    self.data_home = self.default_path if data_home is None else data_home
mirdata/core.py:193: in default_path
    return os.path.join(mir_datasets_dir, self.name)
/usr/lib/python3.8/posixpath.py:90: in join
    genericpath._check_arg_types('join', a, *p)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

funcname = 'join', args = ('/home/miguel/mir_datasets', None), hasstr = True, hasbytes = False, s = None

    def _check_arg_types(funcname, *args):
        hasstr = hasbytes = False
        for s in args:
            if isinstance(s, str):
                hasstr = True
            elif isinstance(s, bytes):
                hasbytes = True
            else:
>               raise TypeError(f'{funcname}() argument must be str, bytes, or '
                                f'os.PathLike object, not {s.__class__.__name__!r}') from None
E               TypeError: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'

/usr/lib/python3.8/genericpath.py:152: TypeError
```

rabitt: Ah my bad! Try:

migperfer: Still not working 😥
I also tried `empty_dataset = core.Dataset(name="empty_test", indexes={"default": 0})`
And then I'm getting `AttributeError: 'int' object has no attribute 'get_path'` not sure how the inner part of `core.Dataset` works.

rabitt: looking into it and will get back to you shortly!

rabitt: @migperfer I just updated the suggested test for test_core - It works for me off of your branch. Let me know if you run into anything though

migperfer: ✨✨Looks like it's working!

rabitt (approved):
Looking great! I'm not sure why black reformatted a bunch of unchanged files... but the check is passing, so let's just go ahead and merge!

PR #532, opened 2021-10-19
vocadito

drubinstein:
Adds vocadito loader and associated tests. Questions to answer before merging:
1. Why is lyrics 13 missing?
2. JAMS is not meant to have multiple notes data as far as I can tell. How should we handle that?
3. For the F0 data what should be used for voicing
4. Lyrics aren't time annotated so is returning a `List[List[str]]` enough?


drubinstein:
@rabitt tests are passing now so you mind re-reviewing?

PR #538, opened 2021-11-06
Patch new splitting functions

### LINKED ISSUE #516: Datasets with train/test/validation splits

### LINKED ISSUE #534: 'Fold' field missing in tinysol
While it is loaded into the metadata, the accessor function for "fold" is missing from the class. It should be added in.

rabitt:
In 0.3.5, we introduced `get_track_splits` and `get_mtrack_splits`, which randomly splits the track/multitrack ids into partitions. One thing we overlooked is that some datasets already come with predetermined splits. In order to support both of these scenarios, this PR:
* renames `get_track_splits` (0.3.5) --> `get_random_track_splits`
* renames `get_mtrack_splits` (0.3.5) --> `get_random_mtrack_splits`
* introduces `get_track_splits`, for datasets with predetermined track splits - these datasets must have a Track.split attribute
* introduces `get_mtrack_splits`, for datasets with predetermined multitrack splits - these datasets must have an MultiTrack.split attribute

I've removed the 0.3.5 release (from yesterday), since we're creating some bad non-backward compatibility here, and will draft a new 0.3.6 release after this change.

Closes #534 
Closes #516 

cc @migperfer @magdalenafuentes


mirdata/core.py
```diff
@@ -295,39 +296,110 @@ def _get_partitions(self, items, splits, seed):
...
 
         Returns:
-            list of lists: a list of lists containing the elements in each split
+            dict: a dictionary containing the elements in each split
         """
-
         if self._track_class is None:
             raise AttributeError("This dataset does not have tracks")
 
         return self._get_partitions(self.track_ids, splits, seed)
```
genisplaja, line 361: Maybe the parameter `split_names` should be passed here as well? If not, it is not really used even if the user sets it.

genisplaja: Same for the multitrack split function!

mirdata/datasets/irmas.py
```diff
@@ -200,49 +201,31 @@ def __init__(
...
+            "dru" in self._audio_filename or "nod" in self._audio_filename
+        ):
+            self.genre = self._audio_filename.split(".")[0].split("[")[2].split("]")[0]
+            self.drum = None
         else:
-            self.train = False
+            self.genre = None
+            self.drum = None
+
+        self.train = True if self.split == "train" else False
```
genisplaja, line 228: Is this really needed if `self.split` is `train`? Looks like redundant info!

genisplaja (approved):

mirdata/datasets/irmas.py
magdalenafuentes: Yes, but removing it is not backwards compatible I guess? We could deprecate it first?

magdalenafuentes (approved):

Issue #541, opened 2021-11-09
Issues with tag_to_jams


bmcfee:
I noticed a couple of small bugs with tag_to_jams (permalink):

https://github.com/mir-dataset-loaders/mirdata/blob/e591c5411c41591e8606812df869dca1ad52ee0f/mirdata/jams_utils.py#L556-L576

1. The documentation copypastad lyrics
2. There is no way to set the confidence value for an observation
3. There is no way to set the time or durations

1 is easy enough to fix.  2 and 3 seem like the same problem, caused by a lack of an underlying tag datastructure (only string is supported).

magdalenafuentes:
Thanks for opening this issue Brian, I'll be working on this and refactoring of jams' converters in #546. One quick check: so far we were treating tags as immutable `str` for the whole track, so I don't get why we need `time` if it's always going to be `time=0.0` (beginning of track). What am I missing?

bmcfee:
Tags can apply to any time interval, so this should be exposed here. It's not a huge deal for most data sets where tags are static, but I see no reason not to support it.

Openmic really does need access to the confidence field though.

PR #544, opened 2021-11-09
[WIP] OpenMIC2018

### LINKED ISSUE #253: OpenMic2018
https://zenodo.org/record/1432913#.XqAkvdMzZ24
https://github.com/cosmir/openmic-2018

bmcfee:
This PR adds a dataset class for openmic2018.  Fixes #253 

The basic functionality is there, but there are a few things left to sort out:

1. How should we handle the individual (anonymized, disaggregated) annotations?
2. JAMS conversion is not done yet.  This is currently blocked by #541 
3. More tests (?) -- the basics are here, but I wanted to get feedback on what's here so far before developing further.


bmcfee:
I think I don't understand why the loader tests are failing / what the expected behavior is here:

https://github.com/mir-dataset-loaders/mirdata/blob/e591c5411c41591e8606812df869dca1ad52ee0f/tests/test_loaders.py#L306-L316

(Incidentally, this would be a bit cleaner using a fixture to iterate over the datasets, rather than sticking them all in a loop inside the test.  You could cut down on prints this way and isolate failures.)

bmcfee:
Bumping this one -- lil help?

genisplaja:
Hello @bmcfee! The problem is in `mirdata/datasets/openmic2018.py`, lines 159-160:
```
# -- set the split
self.split = self._track_metadata.get("split") 
```
The problem is in loading the metadata file in the Track class init. The issue could be fixed by converting this `self.split` attribute into a track property just as `artist` or `title` in your loader. Hope this helps!

bmcfee:
Thanks @genisplaja, that's easy enough to fix.  I still don't understand what this test is trying to accomplish though: it seems like it ought to be an intentional failure.

bmcfee:
Might be a good time to revive this one.. is there anything that needs doing from my side here?

bmcfee:
Attempting to bump this one again

bmcfee:
image

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+        self.vggish_path = self.get_path("vggish")
+
+    @property
+    def split(self):
+        """Get the pre-defined split"""
+        return self._track_metadata.get("split")
+
+    @property
+    def instruments(self) -> Dict[str, float]:
+        """The instrments for this track.
```
nkundiushuti, line 165: typo

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+        A score less than 0.5 indicates that the instrument is probably not present.
+
+        Returns:
+            * dict : instrument name -> confidence score
+        """
+
+        scores = dict()
+        for k in INSTRUMENTS:
+            inst_score = self._track_metadata.get(k, np.nan)
+            if np.isnan(inst_score):
```
nkundiushuti, line 180: maybe replace this with an assert?

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+            track_class=Track,
+            bibtex=BIBTEX,
+            indexes=INDEXES,
+            remotes=REMOTES,
+            license_info=LICENSE_INFO,
+        )
+
+    @core.cached_property
+    def _metadata(self):
+        metadata_path = Path(self.data_home) / "openmic-2018-metadata.csv"
```
nkundiushuti, line 308: could this fail on Windows? maybe replace with os.path.join?

nkundiushuti: similar issues below

nkundiushuti (changes requested):
sorry for the delay, Brian. this looks good. I have just added some small questions regarding the way you construct the paths.

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+
+    Args:
+        fhandle (str or file-like): path or file-like object pointing to an audio file
+
+    Returns:
+        * np.ndarray - the audio signal
+        * float - The sample rate of the audio file
+
+    """
+    # -- load as 44100 mono
```
PRamoneda, line 285: I think the comment is not needed

scripts/make_openmic2018_index.py
```diff
@@ -0,0 +1,76 @@
...
+from mirdata.validate import md5
+
+DATASET_INDEX_PATH = "../mirdata/datasets/indexes/openmic2018_index.json"
+
+
+def make_dataset_index(dataset_data_path):
+    path = Path(dataset_data_path)
+
+    # top-key level metadata
+    metadata_checksum = md5(path / Path("openmic-2018-metadata.csv"))
```
PRamoneda, line 18: It is only a code-style comment, but we use `os.path.join(...)` in the rest of the library. But it is the same... I think

setup.py
```diff
@@ -43,6 +43,7 @@
             "h5py >= 2.10.0",
             "smart_open >= 5.0.0",
             "Deprecated >= 1.2.13",
+            "pandas >= 1.0.0",
```
PRamoneda, line 46: remove this dependency. see above

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+    For more details, please visit: https://zenodo.org/record/1432913
+"""
+import json
+import re
+from pathlib import Path
+from typing import BinaryIO, Optional, Tuple, Dict, List
+
+import librosa
+import numpy as np
+import pandas as pd
```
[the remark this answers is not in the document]
PRamoneda: (you said to @magdalenafuentes I think)

mirdata/datasets/openmic2018.py
bmcfee: No, it won't fail on windows.  Pathlib implements `__div__` as syntactic sugar for something like os.path.join, and is generally the more modern way to do it.

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
bmcfee: I don't think this works?  Your suggestion refers to `inst_score` before it is defined.

scripts/make_openmic2018_index.py
bmcfee: yes, it is the same, and a bit cleaner this way IMO.  Happy to change, but my understanding is that pathlib (which is python standard lib now) is the more modern way to do it.

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+  version      = {v1.0.0},
+  doi          = {10.5281/zenodo.1432913},
+  url          = {https://doi.org/10.5281/zenodo.1432913}
+}
+"""
+
+INDEXES = {
+    "default": "1.0",
+    "test": "1.0",
+    "1.0": core.Index(filename="openmic2018_index.json"),
```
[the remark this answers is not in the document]
bmcfee: I can do that - but I'll also note that it is much smaller than several other indices that are already included (datacos, slakh, good_sounds).  Is there a reason that those are bundled despite their size?  Would it make sense to push them all to remote indices as a separate PR?

mirdata/datasets/openmic2018.py
```diff
@@ -0,0 +1,370 @@
...
+            * np.ndarray - VGGish features; shape=(n_frames, 128)
+        """
+
+        with open(self.vggish_path, "r") as fd:
+            data = json.load(fd)
+            times = np.asarray(data["time_points"])
+            vgg = np.asarray(data["features"])
+            return times, vgg
+
+    # -- will be fed as beat_data=[(self.beats, None)], see jams_utils), and returns a jams
```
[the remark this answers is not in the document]
bmcfee: It's been so long that I don't remember what exactly the hold-up was here, but I believe it comes from #541 and may be blocked until #546 is merged.

setup.py
bmcfee: You do have this dependency transitively through jams, but I understand if you want to remove direct usage from the package.

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
PRamoneda: jejeje typo. I was referring to removing the `continue`

```
for k in INSTRUMENTS:
            inst_score = self._track_metadata.get(k, np.nan)
            if not np.isnan(inst_score):
                   scores[k] = inst_score
```

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
PRamoneda: the standard library CSV xd but maybe you spend some hours doing the table pivoting. There is another alternative: a partial dependency. See: https://github.com/mir-dataset-loaders/mirdata/blob/e591c5411c41591e8606812df869dca1ad52ee0f/mirdata/datasets/haydn_op20.py#L21 .
What do you think @magdalenafuentes ?

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
bmcfee: ah that makes more sense!  Sure, that's fine.

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
PRamoneda: I think we don't have a threshold. If you think is not larger enough. It is fine!

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
bmcfee: I don't think there's any point to a dynamic dependency here - if jams is installed (a hard dependency) then so is pandas.

Now, at some point we might want to make pandas a soft dependency for jams - really it only gets used for annotation→dataframe conversion, which isn't strictly necessary, but that's not currently how it works.

I used it directly for csv parsing mainly out of laziness (and clarity of implementation) - doing it directly with csv is fine, but takes more code to do the same job.

setup.py
PRamoneda: I think i don't have the authority for this decision 😁

scripts/make_openmic2018_index.py
PRamoneda: perfect for me

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
bmcfee: Yeah - I mean, it's pretty bad style to have megabytes of data wrapped up in a python package.  There was a talk on exactly this subject at scipy this year: https://github.com/jameslamb/talks/tree/main/does-that-csv-belong-on-pypi

All of this stuff should really be dynamically fetched only when asked for.  I use pooch for this in librosa, and it works great, see https://github.com/librosa/librosa/pull/1150 .  But I think it's a separate issue to this PR.

nkundiushuti:
there is an error in one of the existing tests in the master branch indeed. I will look into it this week.

nkundiushuti:
I have fixed the problem on the main branch. so you can update this PR with the master

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
nkundiushuti: I think we can include pandas as a partial dependency, similarly to what we did to dali and haydn

setup.py
nkundiushuti: so do we add this as a partial dependency? if it comes already with jams, then maybe we just include it here.

nkundiushuti: @magdalenafuentes any thoughts on this?

mirdata/datasets/openmic2018.py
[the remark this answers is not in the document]
nkundiushuti: if it's much smaller than datacos and other existing index files let's just include it and we decide later on what we do with this index issue on another PR.

nkundiushuti:
if no one objects, I propose merging this into main branch. pandas is already installed as a dependency of jams

nkundiushuti (approved):

bmcfee:
Thanks for merging this - I do think there are still some unresolved issues to clear up as noted in my original post:

> 1. How should we handle the individual (anonymized, disaggregated) annotations?
> 2. JAMS conversion is not done yet. This is currently blocked by https://github.com/mir-dataset-loaders/mirdata/issues/541
> 4. More tests (?) -- the basics are here, but I wanted to get feedback on what's here so far before developing further.

nkundiushuti:
hi Brian!
could you add 1 and 3 as issues so we keep track of them?

bmcfee:
@nkundiushuti to be honest, this PR languished for so long that I don't have a clear recollection of the issues.

For (3), I don't think I ever fully understood what the tests were, or what else could/should be added.  I think this should be more on the maintainers to decide rather than contributors.

For (1), the question is pretty well described in #253 - I'd suggest re-opening that issue until it's entirely resolved.  There are also some other finnicky points in #253 that I raised for discussion, but never received any feedback on (eg erroring out when users request random splits).

PR #555, opened 2022-08-24
fix broken link for "OMRAS2 Metadata Project 2009"

CyxouD:
A previous link for "OMRAS2 Metadata Project 2009" gives "404 Not found error", I replaced it with the one from the CiteSeerX.


genisplaja:
hey everyone! I actually checked and both the links you provided and the one proposed by @harshpalan work, and they point to the same paper. I am merging master, probably some checks will fail because of problems in master, we will be working on that these days and get this merged ASAP.

genisplaja (approved):
Link works, thanks @CyxouD!

Issue #561, opened 2022-10-26
Folder Name with ":" doesn't not allow the repository to be cloned on Windows


harshpalan:
```
 git clone https://github.com/mir-dataset-loaders/mirdata.git
Cloning into 'mirdata'...
remote: Enumerating objects: 8593, done.
remote: Counting objects: 100% (3154/3154), done.
remote: Compressing objects: 100% (1114/1114), done.
remote: Total 8593 (delta 2211), reused 2848 (delta 1961), pack-reused 5439Receiving objects: 100% (8593/8593), 507.12 MiB | 10.98 MiB/s
Receiving objects: 100% (8593/8593), 510.49 MiB | 8.10 MiB/s, done.
Resolving deltas: 100% (5992/5992), done.
error: invalid path 'tests/resources/mir_datasets/saraga_hindustani/saraga1.5_hindustani/Geetinandan : Part-3 by Ajoy Chakrabarty/Bairagi/Bairagi.bpm-manual.txt'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

Geetinandan: Part-3 by Ajoy Chakrabarty is the Folder that creates this issue. We need to modify this so that the repository is compatible with Windows Users also. We discussed this with @magdalenafuentes. @genisplaja @nkundiushuti @rabitt Let us know your thoughts on this. Thanks.

genisplaja:
Thanks for bringing this up! Sorry that I haven't been using Windows a lot recently. So it seems like `:` is not a valid character for filepaths in Windows, so actually, the dataset itself could create problems in a Windows environment.
Probably the safest solution is that we create a new version of the dataset and remove these problematic characters (we would probably also remove the spaces from filenames). I'll discuss that with the rest of our team en let you know.

In the meantime, I can create a PR using a different testing example for the Saraga Hindustani that does not uses `:` in the name so that mirdata can be successfully cloned and installed in Windows, and maybe even we could warn the user that this dataloader is not working on Windows until a new version is uploaded. What you do think?

magdalenafuentes:
Hey @genisplaja, that solution sounds great to me! I'm curious now if this happens with any other dataset, we should probably do a comprehensive testing of leaders in a Windows machine at some point. For now, the fix you propose will allow for Windows users to contribute, which they couldn't!

genisplaja:
[...]

PR #563, opened 2022-10-28
Use different testing example for Saraga Hindustani

genisplaja:
Temporal fix for #561. In the future, we will update the Saraga collections probably with newer and improved data, and fix this naming issue.


harshpalan (approved):
Thank you so much for the quick fix @genisplaja. Approving the pull request.

Issue #580, opened 2023-03-01
Remove support for Python 3.6, adding 3.9


guillemcortes:
Hi all,

I open this issue as a reference thread for the upcoming PR I'll be doing in which I'll be addressing current mirdata issues (I have @genisplaja green light):
- Fixing black version to 23.1.0
- Bump librosa related stuff to use array (so it doesn't break python3.6 support)
- Fix tox.ini to have the three passenv variables in new lines
- Mock pandas import at docs/conf.py for sphinx autodoc

Path forward:
- Support to python3.9 and python3.10 and terminate python 3.6 support.

guillemcortes:
python3.6 test is failing because black 23.1.0 is not available. Should we terminate python3.6 support in this release?
https://app.circleci.com/pipelines/github/mir-dataset-loaders/mirdata/2651/workflows/8ead522b-aec5-4b06-be1a-3ab8e90c42e3/jobs/9351

genisplaja:
Maybe we could soften the black version for now? Like `black>=22.x.x`, so that in python 3.6 we install the original, and for python 3.7 and 3.8 we install the new one and the tests in circle ci should pass!

guillemcortes:
But if python 3.6 installs black 22.x.x and python 3.7 installs 23.x.x both tests won't pass because you will have formatted the files at one specific black version so when testing with the other the test will fail. The easiest/quickest option would be to fix black version to the latest available in python3.6 and format the files with it. Python3.7 and 3.8 would also install the same version so the tests won't fail. Maybe we can do this and update the black version to >= 23.1.0 when we terminate py36 support.

genisplaja:
oops of course, you're right. Sorry about that :) Yes probably what you suggest is the best option for now, so we can move on with the ongoing PRs, and when we remove support for 3.6 we update black to the 23 version.

guillemcortes:
great, I'll go ahead, then :)

PR #586, opened 2023-04-19
#580 python39 and remove python3.6

guillemcortes:
Reference issue: https://github.com/mir-dataset-loaders/mirdata/issues/580


mirdata/jams_utils.py
```diff
@@ -88,10 +88,10 @@ def jams_converter(
 
         try:
             with open(audio_path, "rb") as fhandle:
-                duration = librosa.get_duration(filename=fhandle)
+                duration = librosa.get_duration(path=fhandle)
         # for local mp3s only
         except TypeError:
-            duration = librosa.get_duration(filename=audio_path)
+            duration = librosa.get_duration(path=audio_path)
```
guillemcortes, line 94: @genisplaja I'm not sure what's the motivation behind reading the audio using `with open()` to get the duration later rather than using librosa.load... Also, not sure why now it's not passing codecov and previously it was.

README.md
```diff
@@ -1,6 +1,7 @@
 # mirdata
 common loaders for Music Information Retrieval (MIR) datasets. Find the API documentation [here](https://mirdata.readthedocs.io/).
 
+[![python](https://img.shields.io/badge/python-3.7%20%7C%203.8%20%7C%203.9-blue)](https://www.python.org/)
```
genisplaja, line 4: nice :)

docs/source/contributing.rst
```diff
@@ -39,9 +39,8 @@ To install ``mirdata`` for development purposes:
 
 
 We recommend to install `pyenv <https://github.com/pyenv/pyenv#installation>`_ to manage your Python versions
-and install all ``mirdata`` requirements. You will want to install the latest versions of Python 3.6 and 3.7.
-Once ``pyenv`` and the Python versions are configured, install ``pytest``. Make sure you installed all the pytest
-plugins to automatically test your code successfully. Finally, run:
+and install all ``mirdata`` requirements. Once ``pyenv`` and the Python versions are configured, install ``pytest``.
```
genisplaja, line 42: are we skipping the sentences "you will want to install the latests version of Python 3.7 and 3.8"?

mirdata/jams_utils.py
genisplaja: How is that standing now? Should we still take a look?

genisplaja (approved):
Thanks @guillemcortes, looks good :) Just left a couple of comments (actually more questions than comments). Just a couple of minor things:

- Should we mention somewhere that we discontinue the support to Python3.6 but that if someone is interested on using that they can get mirdata<=0.3.6 (or 0.3.7 actually) which is the last one compatible with Python3.6?
- Probably that fact that the 3.6 test is still there is because the circleci config right? Maybe we could merge and see what happens in the other open PR?

mirdata/jams_utils.py
guillemcortes: this is ok, now. I don't know what happened the other time I tried, but seems ok now. Actually it is a future deprecation warning from librosa, so good to incorporate that in this PR

guillemcortes:
>     * Should we mention somewhere that we discontinue the support to Python3.6 but that if someone is interested on using that they can get mirdata<=0.3.6 (or 0.3.7 actually) which is the last one compatible with Python3.6?
I think we can include this into the next release notes. If this is too hidden, we can maybe add a changelog.md file with the info for easy tracking?
>     * Probably that fact that the 3.6 test is still there is because the circleci config right? Maybe we could merge and see what happens in the other open PR?
yeah, let's see if @magdalenafuentes  got more info, but yeah, it is strange behavior

docs/source/contributing.rst
genisplaja: maybe we could generalize it to "you will want to install the latest versions of Python"... if we want it to be here of course.

guillemcortes:
Closing the PR since this has ben already addressed in #596

PR #595, opened 2023-10-10
Adding loader for IDMT-SMT-AUDIO-EFFECTS

francescopapaleo:
### Adding loader for IDMT-SMT-AUDIO-EFFECTS dataset

Docstring `idmt_smt_audio_effects.py:`

- [x] Create a script in `scripts/`, e.g. `make_idmt_smt_audio_effects.py`, which generates an index file.
- [x] Run the script on the canonical version of the dataset and save the index in `mirdata/indexes/` e.g. `idmt_smt_audio_effects_index.json`.
- [x] Create a module in mirdata, e.g. `mirdata/idmt_smt_audio_effects.py`
- [x] Create tests for your loader in `tests/datasets/`, e.g. `test_idmt_smt_audio_effects.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/table.rst`

**Dataset is fully downloadable.**


mirdata/datasets/idmt_smt_audio_effects.py
```diff
@@ -0,0 +1,228 @@
...
+        midi_nr (int): midi number of the note
+        string (int): string number of the note
+        fret (int): fret number of the note
+        fx_group (int): effect group number
+        fx_type (int): effect type number
+        fx_setting (int): effect setting number
+
+
+    Cached Properties:
+        annotation (EventData): a description of this annotation
```
genisplaja, line 106: You don't actually have that right?

mirdata/datasets/idmt_smt_audio_effects.py
```diff
@@ -0,0 +1,228 @@
...
+                "fx_type": int,
+                "fx_setting": int,
+            }
+        }
+
+        for root, dirs, files in os.walk(self.data_home):
+            for file in files:
+                if file.endswith(".xml"):
+                    xml_path = os.path.join(root, file)
+                    tree = ET.parse(xml_path)
```
genisplaja, line 209: You would need to add something like that here:
```
try:
    tree = ET.parse(xml_path)
except:
   except FileNotFoundError:
   raise FileNotFoundError("Metadata file for not found. Did you run .download()?")
```
Maybe in your case it would be nice if you add a little stentence in the error message that you get the metadata from multiple files and that if there is one missing the error is gonna show as well.

genisplaja (commented):
Looks good! Adding some suggested improvements to get the tests passing. Haven't run the tests locally myself so maybe other ones will show up, let's further take a look on that then. Another change you'd need to make is in the `setup.py`:
```
"librosa >= 0.8.0"
"numpy>=1.16, <=1.20"
```
Please change the versions of these two dependencies for that above. Well, let's see how it goes from these changes on! Thanks again for contributing to mirdata :)

guillemcortes:
Hi @francescopapaleo , could you try updating your branch with the latest changes in master so the new test will run? Thanks!

mirdata/datasets/idmt_smt_audio_effects.py
```diff
@@ -0,0 +1,287 @@
...
+                            "fx_group": int(fxgroup),
+                            "fx_type": int(fxtype),
+                            "fx_setting": int(fxsetting),
+                        }
+
+        if xml_files_count == 0:
+            raise FileNotFoundError(
+                f"No XML files found in {self.data_home}. Did you run .download?"
+            )
+        return metadata
```
genisplaja, line 287: Cool! Now that I think about it... maybe it'd be nice to check that ALL files are there? Well we are actually throwing this error above:
```
except ET.ParseError:
    raise ValueError(
        f"Error parsing XML file {xml_path}. The file may be corrupted."
    )
```
Maybe with saying here `The file may be corrupted or not available, make sure you have all files.`, it's enough... what do you think @guillemcortes @magdalenafuentes?

mirdata/datasets/idmt_smt_audio_effects.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Attributes:
+        audio_path (str): path to audio file.
+
+    Cached Properties:
+        instrument (str): instrument used to record the track.
+        midi_nr (int): midi number of the note.
+        fx_group (int): effect group number.
+        fx_type (int): effect type number.
+        fx_setting (int): effect setting number.
```
genisplaja, line 109: These are  not `Cached Properties`, right...? I think they should be documented as `Attributes` here.

genisplaja (approved):
Hey @francescopapaleo I am accepting the PR :) Left a few minor comments/questions, if could please update that, in my opinion the loader will be ready to merge! Thanks!!

mirdata/datasets/idmt_smt_audio_effects.py
guillemcortes: Hi,

_(Minor comment)_
Since you are not using `dirs`, maybe I'll change `for root, dirs, files in os.walk(self.data_home):` for `for root, _, files in os.walk(self.data_home):`

Also, the ParseError is only raised if the xml cannot be parsed, but the file exists (os.walk finds it), so I think the message is appropiate. Maybe you can also add something like `The file may be corrupted or not available`.

To check that all the xml have been processed maybe you can add to the final `if xml_files_count == 0` an additional `else if` that checks that `xml_files_count==expected_xml_count` and raises an error if the numbers don't coincide. So you can check that the total amount of xml files that have been loaded is the same amount that is expected.

guillemcortes (approved):

Issue #628, opened 2024-04-26
Ballroom Dataset: genre not available


Masetto96:
The track object for the ballroom dataset contains only annotations for beats and tempo but **it is missing the genre attribute** (e.g. waltz, samba).

https://mirdata.readthedocs.io/en/latest/source/mirdata.html#mirdata.datasets.ballroom.Track

PR #629, opened 2024-04-30
Added genre to ballroom datataset loader

Masetto96:
https://github.com/mir-dataset-loaders/mirdata/issues/628

The track object of the ballroom dataset does not have the annotation about the genre of the track. Therefore, the genre attribute have been included from the name of the audio path.

Test for getting the genre from the audio path has been implemented.

Following the contribution guidelines, code has been formatted and all tests passed.


Masetto96:
@genisplaja @guillemcortes plz review :)

genisplaja:
hey @Masetto96, I did finally take a look at this one and LGTM. Let us update #630 so that codecov works as expected and we can try to merge this one.

mirdata/datasets/ballroom.py
```diff
@@ -101,10 +101,12 @@ class Track(core.Track):
         audio_path (str): path to audio file
         beats_path (str): path to beats file
         tempo_path (str): path to tempo file
+        genre (str): genre of the track
 
     Cached Properties:
         beats (BeatData): human-labeled beat annotations
         tempo (float): human-labeled tempo annotations
+
```
genisplaja, line 109: remove added enter. Or is it because formatting issues?

mirdata/datasets/ballroom.py
```diff
@@ -164,6 +168,20 @@ def to_jams(self):
         )
 
 
+def get_genre(audio_path: Optional[str]) -> Optional[str]:
```
genisplaja, line 171: normally, for consistency, we name this functions `load_<thing>()`. I think it would be nice if you could use this convention here as well

genisplaja (changes requested):
Right! Almost there, I guess it works as expected, and I don't see any missing tests. Just requesting minor changes for consistency and that's ready :)

mirdata/datasets/ballroom.py
Masetto96: Removed the new line, however I noticed that in other scripts it is sometimes present

mirdata/datasets/ballroom.py
```diff
@@ -130,6 +132,8 @@ def __init__(
         self.beats_path = self.get_path("beats")
         self.tempo_path = self.get_path("tempo")
 
+        self.genre = get_genre(self.audio_path)
```
Masetto96, line 134: While making the changes I spotted a mistake. Here I changed line 135 to self.load_genre()

tests/datasets/test_ballroom.py
```diff
@@ -94,3 +96,12 @@ def test_load_audio():
     assert type(audio) == np.ndarray
 
     assert ballroom.load_audio(None) is None
+
+
+def test_get_genre():
+    data_home = "tests/resources/mir_datasets/ballroom"
+    dataset = ballroom.Dataset(data_home)
+    track = dataset.track("Media-105901")
+    genre = ballroom.get_genre(track.audio_path)
```
Masetto96, line 105: Made necessary changes also here

mirdata/datasets/ballroom.py
genisplaja: Ok! Was just not to unnecessary change lines. I guess it does not affect at all, I can take a look

mirdata/datasets/ballroom.py
genisplaja: Yes! Although are you sure is `self.load_genre()`? I think that the `load_genre()` function is outside the class right? Shouldn't it be without the `self.`?

Masetto96: Oops, you are right! I assumed it was part of the class but I did not look carefully. Sorry about that!

genisplaja (approved):
LGTM! I guess we need PR #631 to be merged so that the tests pass, and then we can propagate the update to this PR and merge. Thanks @Masetto96 :)

genisplaja:
hey @Masetto96, very close but there is a small thing that we are still missing. All custom `load_*` methods are passed through a test to check that if a wrong filepath is given, an error should be thrown, but in your case, the function returns `None`. I guess all files should have a genre right? Because all audio files are contained in a "genre"-named folder. Then it would make sense that an error is thrown if the audio path is not found. 

One thing you could do is use the decorator as other loading functions. See the load_audio below:

```
@io.coerce_to_bytes_io
def load_audio(fhandle: BinaryIO) -> Tuple[np.ndarray, float]:
    """Load a Ballroom audio file.
    Args:
        fhandle (str or file-like): path or file-like object pointing to an audio file
    Returns:
        * np.ndarray - the mono audio signal
        * float - The sample rate of the audio file
    """
    return librosa.load(fhandle, sr=None, mono=True)
```

If you use that decorator, you will pass directly a file handle and if this file is not found, the expected error by the core functions in the library will be thrown. Let me know if that was clear!

Masetto96:
Done! Let me know if I there is something else. Thank you for the code review.

genisplaja:
Hey @Masetto96 man, finally fetched your fork and played around to figure out the issue. Sorry that I didn't notice that we were using a too overcomplicated solution for a problem that should be way easier to address. Let me explain how we should be actually loading the genre:

- The `load_genre` method should not exist.
- You can directly get `self.genre` as an attribute of the `Track` class like this: `self.genre = os.path.basename(os.path.dirname(self.audio_path)).lower()`, in the `Track` init.
- Then, the `test_load_genre` is also not needed anymore.

In the end, the genre annotation is not contained in a local file, but rather in a filepath that is already contained in the library itself (in the .json index), so you wouldn't even need to download the dataset to get the genre. Therefore, we don't need a cached property for that. `genre` can be a simple attribute of the ballroom `Track` class and you would read from the path. Sorry that I did not notice that earlier... thanks for following my suggestions man. If you could update that, I am pretty sure the tests will pass (are passing locally for me), so we will be able to move on with that one. Thanks again and sorry for the confusion :)

Masetto96:
Of course, your suggestion makes a lot of sense!

PR #645, opened 2024-12-01
Fix wrong Carnatic Music Rhythm Zenodo link

genisplaja:
Just spotted a wrong link for the index remotes in the Carnatic Music Rhythm dataset. This PR fixes it. 

That is the misspelling: hhttps --> https


guillemcortes (approved):
Good catch :)

PR #649, opened 2024-12-12
Adding loader for MIR-1K

dsuedholt:
### Description

### Dataset loaders checklist:

- [x] Run the script on the canonical version of the dataset and upload the index to [Zenodo Audio Data Loaders community](https://zenodo.org/communities/audio-data-loaders).
- [x] Create a sample version of the index with the necessary information for testing.
- [x] Create tests for your loader in `tests/datasets/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/table.rst`
- [x] Run `black`, `flake8` and `mypy` (see [Running your tests locally](https://mirdata.readthedocs.io/en/stable/source/contributing.html?highlight=contributing#running-your-tests-locally)).
- [x] Check that codecov coverage does not decrease.

### Other recordings

The MIR-1K dataset is based on 110 song recordings split into 1000 clips. 

The dataset download also includes a folder `UndividedWavfile` which contains the "un-split" recordings. Additionally, while all other annotation files are on the clip level, the download also includes a `LyricsWav` folder containing recordings of the song lyrics being read aloud (as speech, not singing), but these correspond to the 110 "un-split" songs, not the 1000 clips. So I wasn't quite sure how to handle these within the mirdata structures, for now I simply ignored them.


tests/test_full_dataset.py
```diff
@@ -92,7 +92,7 @@ def test_load_mtracks(skip_remote, dataset):
 
     assert isinstance(all_data, dict)
 
-    mtrack_ids = dataset.mm
+    mtrack_ids = dataset.mtrack_ids
```
genisplaja, line 95: Good catch, and sorry about that. We are going to be adding a fix for that in a separate PR here: #657.

mirdata/datasets/mir_1k.py
```diff
@@ -0,0 +1,323 @@
...
+    Raises:
+        IOError: if lyrics_path does not exist
+
+    Returns:
+        str: the lyrics
+
+    """
+
+    return annotations.LyricData(
+        intervals=np.array([[0.0, 0.0]]),
```
genisplaja, line 258: I was wondering if it was worth getting the duration of the track and having an interval like `[0.0, 10.0]`, `10.0` being an example total duration of the track. Maybe the faster way would be to parse it from the annotations. However, I wonder if there's chance that someone needs lyrics+duration and no audio/annotations from where to get the duration also.

mirdata/datasets/mir_1k.py
```diff
@@ -0,0 +1,323 @@
...
+        times=times,
+        time_unit="s",
+        frequencies=f0,
+        frequency_unit="midi",
+        voicing=(f0 > 0).astype(np.float64),
+        voicing_unit="binary",
+    )
+
+
+# not using @io.coerce_to_string_io here because it has hardcoded utf-8 encoding
```
genisplaja, line 242: Interesting, noting this issue. We could have a tailored decorator later.

mirdata/datasets/mir_1k.py
```diff
@@ -0,0 +1,323 @@
...
+        audio_path (str): path to the track's audio file
+        f0_path (str): path to the track's f0 annotation file
+        lyrics_path (str): path to the track's lyrics file
+        track_id (str): track id
+        unvoiced_labels_path (str): path to the track's unvoiced labels
+        vocal_activity_path (str): path to the track's vocal activity labels
+
+    Cached Properties:
+        f0 (F0Data): the track's f0 annotation
+        lyrics (LyricData): the track's lyrics
```
genisplaja, line 93: maybe if we end up leaving the lyrics with no interval/duration information, we could also let the user know here?

genisplaja (commented):
Hey @dsuedholt, thanks for the contribution and sorry for the late review. I am leaving a few questions before accepting the PR, to hear your opinons and potentially opinions from peer maintainers. 

Also, should be ignore the non-split portion of the dataset? Is it normally used or not used in related research works? I guess loading them would mean have a bunch of more tracks included in the dataset, but those would mostly be already present in the dataset in the split format right? 

Thanks again for your work, I did also take a look at the upload request for the index in Zenodo and looks fine, so I am going to be accepting it :)

mirdata/datasets/mir_1k.py
dsuedholt: Good point - potentially this could also be a way to include the lyric data annotation with the unsplit files. i.e. assume a track is split into clips of 10 second lengths, then for the unsplit track, include the lyrics annotations for each clip with the intervals `[0, 10], [10, 20]` etc?

magdalenafuentes:
Hey @dsuedholt, thanks for this PR, it looks great! I have just one comment/suggestion. I noticed that you are using the `Track` object with `instrumental` and `vocal` audio. I would suggest that you use the [`MultiTrack`](https://github.com/mir-dataset-loaders/mirdata/blob/master/mirdata/datasets/jtd.py#L283) object instead. My concern is that from the user's perspective, `Track` is expected to have a single attribute `audio` and it might be confusing, whereas  `MultiTrack` is more flexible and can point to different audio by design (e.g. vocals, instrumental). Thoughts?

PR #654, opened 2025-02-13
[BREAKING] Removing JAMS and python update

### LINKED ISSUE #623: ModuleNotFoundError: No module named 'smart_open'
2024 Macbook Pro M3 Max
MacOS Sonoma 14.1
Python 3.10.4 via pyenv virtualenv

I installed mirdata via pip through a pyproject.toml file, and after installation, calling mirdata.initialize for any dataset generates the following error message:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "$HOME/.pyenv/versions/{pyenv-virtualenv-name}/lib/python3.10/site-packages/mirdata/__init__.py", line 53, in initialize
    module = importlib.import_module("mirdata.datasets.{}".format(dataset_name))
  File "/Users/benjamingenchel/.pyenv/versions/3.10.4/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "$HOME/.pyenv/versions/{pyenv-virtualenv-name}/lib/python3.10/site-packages/mirdata/datasets/slakh.py", line 35, in <module>
    from smart_open import open
ModuleNotFoundError: No module named 'smart_open'
```

I can install it separately, after which the call works, but that doesn't seem right. If it's necessary for the package to run it should be installed as a dependency.

### LINKED ISSUE #627: Python 3.12 support
It looks like mirdata still depends on imp, which was deprecated in 3.11 and removed in 3.12:

```
pip install mirdata 
Collecting mirdata
  Using cached mirdata-0.3.8-py3-none-any.whl.metadata (6.6 kB)
Collecting attrs>=23.1.0 (from mirdata)
  Using cached attrs-23.2.0-py3-none-any.whl.metadata (9.5 kB)
Collecting black>=23.3.0 (from mirdata)
  Using cached black-24.3.0-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (75 kB)
Collecting chardet>=5.0.0 (from mirdata)
  Using cached chardet-5.2.0-py3-none-any.whl.metadata (3.4 kB)
Collecting Deprecated>=1.2.14 (from mirdata)
  Using cached Deprecated-1.2.14-py2.py3-none-any.whl.metadata (5.4 kB)
Collecting jams>=0.3.4 (from mirdata)
  Using cached jams-0.3.4.tar.gz (51 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-gavk7r0k/jams_b471981ad2754869a2455feefe6ac42e/setup.py", line 3, in <module>
          import imp
      ModuleNotFoundError: No module named 'imp'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
```

Recommend to replace imp with importlib: https://docs.python.org/3/library/importlib.html#module-importlib

### LINKED ISSUE #653: GuitarSet loading is slow
Loading guitarset is quite slow. Extracting `multif0` and `note_all` both take over 2s per file, thats >24minutes to load the whole dataset, which, frankly, is not that big.

From what I see both of these attributes each call 6x a load routine that seems to parse the whole jams file, resulting in 6 parsings for each attribute, so the jams file is parsed 12 times. I'm no expert in how jams files are parsed but seeing that the `jams.load(fhandle)` takes about 80% of the time of `load_pitch_contour`, I'm assuming this could be significantly faster.

Options for improvement:
- make the contents of the jams file a cached attribute 
-> best option if memory is not an issue, given it's "just" 540 MB it's not the worst ever (although python probably blows that up..)
- extend routines like `load_pitch_contour` to accept either the `jams_path` or the contents of the file (`jam`). This way the `multif0` (and similar) attributes can load the file once and pass it on 6 times.
-> avoids keeping unnecessary memory, but still reads the file multiple times when accessing different attributes
-  same as above plus add a routine that directly extracts multiple attributes using the same single-read file contents (or reads all attributes and the ones discarded are then freed up again) 
-> best of both but requires most code changes

Masetto96:
⚠️ [THIS PR INTRODUCES A BREAKING CHANGE IN THE API AS IT DROPS `to_jams()` FROM THE TRACK CLASS]

Dropping JAMS reduces the project's dependencies.

fixes #627, fixes #647, closes #623, resolves #653  
 
1. Removed JAMS dependency from pyproject.toml and yml files of github actions
2. Remove functions, imports, tests and utils related to JAMS
3. Replaced JAMS functionalities with JSON where needed (e.g namespace.search)
4. Updated tests to ensure they all pass with the new JSON implementation
5. Updated and minor fixes to documentation (added button to copy code block) 
6. Added JAMS in tutorial section of docs - how to use jams externally to load mirdata's annotation
7. Added tests in guitarset to fix code coverage
8. Update python support til 3.12 and dropped 3.8

Sidenote: when running the github action for continuos integration I noticed that one test fails because mir_eval is missing. In fact test_annotations uses mir_eval. For this reason I have added mir_eval to environment-ci.yml and pyproject.toml

Edit: mir_eval was a dependency of jams


Masetto96:
@magdalenafuentes fixed the last few things, this should be good to go as well :)

magdalenafuentes (approved):
@Masetto96 fantastic PR, thanks for the great work

Issue #659, opened 2025-03-06
Broken download link for GTZAN Genre Dataset ("gtzan_genre")


fergarciadlc:
**Issue:** Broken download link for GTZAN Genre Dataset (`gtzan_genre`)

### Description
When using `mirdata` to download the `gtzan_genre` dataset, the download fails due to a broken or unavailable link.

### Steps to Reproduce
```python
import mirdata

dataset = mirdata.initialize("gtzan_genre")
dataset.download()
dataset.validate()
```

### Error Message
```
INFO: Downloading ['all', 'tempo_beat_annotations'] to /Users/fernando/mir_datasets/gtzan_genre
INFO: [all] downloading genres.tar.gz
0.00B [00:00, ?B/s]ERROR:
    mirdata failed to download the dataset from http://opihi.cs.uvic.ca/sound/genres.tar.gz!
    Please try again in a few minutes.
    If this error persists, please raise an issue at
    https://github.com/mir-dataset-loaders/mirdata,
    and tag it with 'broken-link'.
```

**Traceback:** (ConnectionResetError)

### Temporary Workaround
I found a manual workaround:
1. Download the missing JSON index file from [Zenodo](https://zenodo.org/records/14004436/files/gtzan_genre_1.0_mini_index.json?download=1).
2. Place the file in:
```
/path-to-env/lib/python3.X/site-packages/mirdata/datasets/indexes/
```

After manually adding the JSON file, the dataset validates and works locally without issues. However, automated tests/pipelines still fail due to the broken download link.

### Additional Information
- Dataset validation works locally after the workaround.
- Automated CI pipelines continue failing due to link issues.

**Suggested Fix:**
Update or mirror the dataset to a stable URL.

---

**Tag:** `broken-link`

Masetto96:
Thank you for bringing this issue to our attention. 

It appears that the zip file for the default version of the dataset is not available, and this issue is unrelated to the missing index. While the files for the mini version are downloadable, the indexes fail to download.

We are currently looking into this and we will provide an update soon.

giovana-morais:
GTZAN is not public anymore (see #549). I'm think the fix for this would be to remove the download link and update the documentation.

Masetto96:
As @giovana-morais mentioned, the default version of GTZAN is no longer available. Thank you for the heads-up. We will update the documentation to reflect this change.

[...]

PR #662, opened 2025-03-28
Bug-Fix downloading index

Masetto96:
This PR fixes https://github.com/mir-dataset-loaders/mirdata/issues/659, a bug where the index json file is not downloaded.
In certain cases, when calling `dataset.download()` the index file is not downloaded as expected.

### Bug Description:
- When `partial_download` is not `None`, `objs_to_download` is set to `partial_download`, which does not contain `"index"`.
- This causes the index to be missing from the objects to be downloaded and therefore it is not downloaded.
- I believe this bug is affecting all the cases in which  `partial_download` is specified.

### Fix:
- Explicitly append `"index"` to `objs_to_download` when `partial_download` is not `None`.


genisplaja (approved):
Nice fix! Tested it locally and works like a charm. Thanks!

magdalenafuentes (approved):

PR #652, opened 2024-12-23
Adding loader for Jazz Trio Database

HuwCheston:
### Description

Please include the following information at the top level docstring for the dataset's module mydataset.py:

- [x] Describe annotations included in the dataset
- [x] Indicate the size of the datasets (e.g. number files and duration, hours)
- [x] Mention the origin of the dataset (e.g. creator, institution)
- [x] Describe the type of music included in the dataset
- [x] Indicate any relevant papers related to the dataset
- [x] Include a description about how the data can be accessed and the license it uses (if applicable) 

#### Dataset loaders checklist:

- [x] Create a script in `scripts/`, e.g. `make_my_dataset_index.py`, which generates an index file.
- [x] Run the script on the canonical version of the dataset and upload the index to [Zenodo Audio Data Loaders community](https://zenodo.org/communities/audio-data-loaders).
    - Currently awaiting approval.
- [x] Create a sample version of the index with the necessary information for testing.
- [x] Create a module in mirdata, e.g. `mirdata/my_dataset.py`
- [x] Create tests for your loader in `tests/datasets/`, e.g. `test_my_dataset.py`
- [x] Add your module to `docs/source/mirdata.rst` and `docs/source/table.rst`
- [x] Run `black`, `flake8` and `mypy` (see [Running your tests locally](https://mirdata.readthedocs.io/en/stable/source/contributing.html?highlight=contributing#running-your-tests-locally)).
- [x] Run `tests/test_full_dataset.py` on your dataset.
    - I have run this test on the canonical version of the database. However, note that I had to make changes to the `Dataset` class in order to pass the `test_load_mtracks` function. See my issue open here #651  
- [x] Check that codecov coverage does not decrease.

If your dataset **is not fully downloadable** there are two extra steps you should follow:
- [x] Contacting the mirdata organizers by opening an issue or PR so we can discuss how to proceed with the closed dataset.
     - I am opening this PR, hopefully this is ok for discussion 😀 
- [x] Show that the version used to create the checksum is the "canonical" one, either by getting the version from the dataset creator, or by verifying equivalence with several other copies of the dataset.
    - I am the creator of the database and have created the checksums using my own, canonical, version.
- [x] Make sure someone has run `pytest -s tests/test_full_dataset.py --local --dataset my_dataset` once on your dataset locally and confirmed it passes.
    - I have run this test on the canonical version of the database. However, note that I had to make changes to the `Dataset` class in order to pass the `test_load_mtracks` function. See my issue open here #651

PR #664, opened 2025-04-02
Add support for `*args` and `**kwargs` in `io.coerce_to_*` decorators

HuwCheston:
As discussed with @genisplaja in https://github.com/mir-dataset-loaders/mirdata/pull/652#pullrequestreview-2735847747, this PR makes an initial attempt to allow for the use of optional `args` and `kwargs` in the two decorator functions inside `io`.

The changes to the decorators themselves are very minor. I also changed a few functions in several loaders that weren't previously using the decorators. These are the same as defined in `test_loaders.EXCEPTIONS`, i.e.:

```
EXCEPTIONS = {
    "dali": {"load_annotations_granularity": {"granularity": "notes"}},
    "guitarset": {
        "load_pitch_contour": {"string_num": 1},
        "load_notes": {"string_num": 1},
        "load_chords": {"leadsheet_version": False},
    },
    "tonas": {"load_f0": {"corrected": True}},
    "compmusic_carnatic_varnam": {
        "load_notation": {
            "taala_path": "a/fake/path",
            "structure_path": "a/fake/path",
        },
    },
}
```

A few other things to note:
- I had to simplify the typehints for both `coerce_to*` functions in order to get `mypy` passing successfully. Someone with more familiarity with this package might have a better idea for how to keep the original type hints in place as much as possible.
- Weirdly, I also had to make a number of changes to functionality in `compmusic_carnatic_varnam.load_notation` for the same `mypy` reasons. Probably the most substantial is replacing calls to `np.arange` with the builtin `range` argument. I don't think this should lead to any functional changes (and tbh I have no idea why `mypy` started throwing additional errors when adding the decorator), but would appreciate someone with more familiarity taking a look.
- There are no changes to any of the tests, all seem to pass successfully on my local machine with the decorator changes made.
- There's also a bonus fix to `cipy.difficulty_annotation` that was throwing up (unrelated) `mypy` errors for me.

### DIFF
--- a/mirdata/datasets/cipi.py
+++ b/mirdata/datasets/cipi.py
@@ -147,7 +147,7 @@ class Track(core.Track):
         )
 
     @property
-    def difficulty_annotation(self) -> int:
+    def difficulty_annotation(self) -> Optional[int]:
         return (
             self._track_metadata["henle"] if "henle" in self._track_metadata else None
         )
--- a/mirdata/datasets/compmusic_carnatic_varnam.py
+++ b/mirdata/datasets/compmusic_carnatic_varnam.py
@@ -41,6 +41,7 @@ import os
 import csv
 import glob
 import librosa
+from typing import TextIO
 
 import numpy as np
 from xml.dom import minidom
@@ -240,8 +241,8 @@ def load_taala(fhandle):
     )
 
 
-# no decorator here because we need three paths
-def load_notation(note_path, taala_path, structure_path):
+@io.coerce_to_string_io
+def load_notation(note_path: TextIO, taala_path: str, structure_path: str):
     """Load notation and structure
 
     Args:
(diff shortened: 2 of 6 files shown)


HuwCheston:
In https://github.com/mir-dataset-loaders/mirdata/pull/664/commits/80bafb9308f3375ed47ddb2e12425f8c8ab7e54e I'm removing a few redundant `try ... except FileNotFound` blocks which should solve the coverage issue in the previous commits ^^

Previously these were necessary before adding the `coerce_to_*` decorators, but now the `FileNotFoundError` will be triggered inside the decorator instead. This was causing the lines within the `except` clause to not be hit when running the tests.

genisplaja (approved):
Hey @HuwCheston, that one looks great as well :) Thanks for that! Tested also locally and everything seems to work fine. I think this is ready to go.

HuwCheston:
nice! feel free to merge whenever you're ready

PR #702, opened 2026-05-25
Fix numpy 2.4+ scalar conversion in gtzan_genre.load_tempo (closes #701)

### LINKED ISSUE #701: Unable to access tempo from gtzan on recent numpy
Numpy >=2.4 does not allow direct casting of arrays with size=1 to scalars.  See https://numpy.org/doc/stable/release/2.4.0-notes.html#raise-typeerror-on-attempt-to-convert-array-with-ndim-0-to-scalar

This causes a failure to access the tempo field of gtzan tracks here:

https://github.com/mir-dataset-loaders/mirdata/blob/b95bf3893616ff16082beed7869f144bb711c163/mirdata/datasets/gtzan_genre.py#L147-L160

One potential fix is to explicitly access the 0th element of the loaded array.  I've hacked this locally and it works:

```
    return float(tempo.flat[0])
```

Probably there should be a thorough audit for other occurrences of this problem across datasets.

mnlscn:
## Summary

Fixes #701. numpy >= 2.4 no longer allows implicit conversion of arrays with `ndim > 0` to Python scalars ([release notes](https://numpy.org/doc/stable/release/2.4.0-notes.html#raise-typeerror-on-attempt-to-convert-array-with-ndim-0-to-scalar)):

 `TypeError: only 0-dimensional arrays can be converted to Python scalars `

`load_tempo` builds a guaranteed `(1, 1)` array via `np.loadtxt(fhandle, ndmin=2)`, so `float(tempo)` now fails. Fix: `float(tempo.flat[0])`, as suggested by @bmcfee in the issue

## Changes

- `mirdata/datasets/gtzan_genre.py`: `float(tempo)` → `float(tempo.flat[0])`

## Audit methodology

Searched the codebase for all `float()` / `int()` calls on numpy array results, and all `np.loadtxt` / `np.genfromtxt` outputs. The distinguishing factor for this bug is `ndmin=2`, which forces a 2-d array even for a single value. Every other loader either *slices* the result (`arr[:, 0]` — returned as an array) or *iterates* it (`for row in arr` — each `row[i]` is a numpy scalar with `ndim=0`, which still converts fine). The thousands of `float(line[i])` calls elsewhere operate on `csv.reader` strings, not arrays

Confirmed empirically: running the full test suite under numpy 2.4.6 produces exactly two `TypeError: only 0-dimensional arrays...` failures, both in `test_gtzan_genre`

## Testing

- Both previously failing tests (`test_track`, `test_load_tempo`) now pass.
- Full suite under numpy 2.4.6: 19 → 17 failed, 342 → 344 passed. No regressions. The remaining 17 failures are all pre-existing and unrelated to this change (optional-dep imports, a Python 3.11+ csv-delimiter regression in two other datasets, and a unicode-zip assertion)

## Notes

I used `.flat[0]` to match @bmcfee's suggestion in the issue. Happy to switch to `.item()` if preferred, it's marginally more defensive (raises `ValueError` on a multi-element file rather than silently returning the first element), but tempo annotation files are single-value by format so the distinction is just academic here I'd say

### DIFF
--- a/docs/source/table.rst
+++ b/docs/source/table.rst
@@ -303,7 +303,7 @@
           :target: https://creativecommons.org/licenses/by-sa/4.0
 
    * - Gtzan-Genre
-     - - audio: ✅
+     - - audio: ❌
        - annotations: ✅
      - - global :ref:`genre`
        - :ref:`beats`
--- a/mirdata/datasets/gtzan_genre.py
+++ b/mirdata/datasets/gtzan_genre.py
@@ -157,7 +157,7 @@ def load_tempo(fhandle: TextIO) -> float:
 
     tempo = np.loadtxt(fhandle, ndmin=2)
 
-    return float(tempo)
+    return float(tempo.flat[0])
 
 
 @io.coerce_to_bytes_io
--- a/mirdata/datasets/gtzan_genre.py
+++ b/mirdata/datasets/gtzan_genre.py
@@ -38,7 +38,7 @@ INDEXES = {
         filename="gtzan_genre_index_1.0.json",
         url="https://zenodo.org/records/13993311/files/gtzan_genre_index_1.0.json?download=1",
         checksum="533ca050855f22acf2feb283d9957fe3",
-        partial_download=["all", "tempo_beat_annotations"],
+        partial_download=["tempo_beat_annotations"],
     ),
     "mini": core.Index(
         filename="gtzan_genre_1.0_mini_index.json",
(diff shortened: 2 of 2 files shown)


yujin-kimmm:
Hi @mnlscn, Thank you for this PR!
Currently, CI won't pass since we have some issue and we are working on it. Once it's ready, we will run the test. Thank you!

mnlscn:
Good to know!
Thanks and good luck with the bug fix :)

magdalenafuentes (approved):

mnlscn:
I resolved the conflict by keeping the current `master` implementation:

```python
return float(np.asarray(tempo).item())
```

This matches the more defensive approach discussed here and makes this PR effectively empty now (`Files changed: 0`), since the fix has already landed on `master`.

Given that, I think this PR can be closed as superseded by the fix already merged into `master`. Thanks again for the review and for taking care of this!

yujin-kimmm:
Hi @mnlscn, 

Sorry for the confusion! this was totally my mistake. I just checked that I made the changes few months ago in #700 and didn't explicitly add the note in the PR, which made me to forget it.

Now that #700 is merged, this issue is fixed. Sorry again for confusion, and thank you again for making this PR!

